PR #334, opened 2024-11-05
feat (coprocessor-node): sp1 support

### LINKED ISSUE #120: coprocessor-node: Add back in sp1
We decided to remove sp1 because it significantly slowed down the build times. We will likely want to add it back in prior to launching, so this issue exists to track that

- commit where sp1 was commented out: https://github.com/Ethos-Works/InfinityVM/commit/7fbb708c7f7506b260d4c9fc15ff8227c5e5cf4b
    - I aimed to annotate every section of commented out code with `https://github.com/Ethos-Works/InfinityVM/issues/120` so we could easily find it when we want to bring it back
- telegram message where sp1 team explains fixes that we should watch for: https://t.me/succinct_sp1/2920
    - one of the PRs we are waiting for: https://github.com/succinctlabs/sp1/pull/1177
    - we also are waiting for them to feature gate the sdk so we can just get the execute without prover. No issue yet
    
Tasks
- [x] Add back sp1 support and the mock consumer program so we can unit test it
- [ ] Create rough comparison with the load test tool https://github.com/InfinityVM/InfinityVM/issues/285

mayuridev:
PR to add back SP1 Support (#120). Currently using SP1 w/ prover client, in future iterations will look into incorporating partitioned executor crate (https://github.com/succinctlabs/sp1/tree/dev/crates/core/executor). 

Will also work on load testing + comparison between Risc0 and SP1. Currently SP1 zkVM program is just a mock consumer, but will incorporate different levels of intensity testing / hashing. 

closes #120


crates/zkvm/src/lib.rs
```diff
@@ -1,7 +1,9 @@
 //! ZKVM trait and implementations. The trait should abstract over any complexities to specific VMs.
 
 use risc0_binfmt::compute_image_id;
-use risc0_zkvm::{Executor, ExecutorEnv, LocalProver};
+use risc0_zkvm::{Executor as Risc0Executor, ExecutorEnv, LocalProver};
+// use sp1_core_executor::{Executor as Sp1CoreExecutor, Program};
```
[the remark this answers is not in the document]
mayuridev: yeah, will open up a branch that only uses executor

programs/sp1/sp1-consumer/script/src/bin/main.rs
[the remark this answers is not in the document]
mayuridev: agree

crates/zkvm/src/lib.rs
```diff
@@ -167,4 +171,15 @@ mod test {
 
         assert!(!correct);
     }
+    #[test]
+    fn sp1_execute_can_correctly_execute_program() {
```
emostov, line 175: This test looks good. Can we also add a test that looks like `risc0_is_correct_verifying_key`

programs/sp1/mock-consumer/script/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "mock-consumer-sp1-script"
+version = "0.1.0"
+edition = "2021"
+
+[build-dependencies]
+sp1-helper = { workspace = true }
+sp1-build = "3.1.0"
```
emostov, line 8: lets use the workspace dep here too

elf/riscv32im-succinct-zkvm-elf
emostov, line 1: i would like this elf folder to be in something like `programs/sp1/mock-consumer` so its clear that its the mock consumer elf. Looking into how to do this rn

Makefile
```diff
@@ -1,5 +1,4 @@
 # These where required for building and running tests with sp1
-# TODO: https://github.com/InfinityVM/InfinityVM/issues/120
 # RUSTFLAGS = '-Copt-level=3 -Cdebug-assertions -Coverflow-checks=y -Cdebuginfo=0 -C target-cpu=native'
```
emostov, line 2: lets also remove the rustflags

Cargo.toml
```diff
@@ -228,3 +230,4 @@ test-utils = { path = "crates/sdk/test-utils" }
 zkvm = { path = "crates/zkvm" }
 zkvm-tracing = { package = "tracing", path = "crates/tracing" }
 zkvm-executor = { path = "crates/zkvm-executor" }
+mock-consumer-sp1 = { path = "programs/sp1/mock-consumer" }
```
MaanavKhaitan, line 233: nit: if we are naming this `mock-consumer-sp1` then can we rename `mock-consumer-methods` to `mock-consumer-risc0`

crates/scripts/src/local.rs
```diff
@@ -10,7 +10,8 @@ use clob_test_utils::{anvil_with_clob_consumer, mint_and_approve};
 use contracts::{DeployInfo, DEFAULT_DEPLOY_INFO};
 use intensity_test_methods::INTENSITY_TEST_GUEST_ELF;
 use mock_consumer::anvil_with_mock_consumer;
-use mock_consumer_methods::MOCK_CONSUMER_GUEST_ELF;
+use mock_consumer_methods::MOCK_CONSUMER_RISC0_GUEST_ELF;
+use mock_consumer_sp1::SP1_MOCK_CONSUMER_GUEST_ELF;
```
MaanavKhaitan, line 14: nit: let's stay consistent with naming. if it's `MOCK_CONSUMER_RISC0_GUEST_ELF` then it should also be `MOCK_CONSUMER_SP1_GUEST_ELF`

elf/riscv32im-succinct-zkvm-elf
MaanavKhaitan: yep I agree

crates/sdk/proto/src/coprocessor_node.v1.rs
```diff
@@ -204,6 +204,7 @@ pub struct SubmitProgramResponse {
 pub enum VmType {
     Risc0 = 0,
     Sp1 = 1,
```
MaanavKhaitan, line 206: do we have any benchmarks on performance and compile times of SP1 vs Risc0?

If SP1 is better or even the same as Risc0, I'd be in favor of just removing Risc0 entirely from the codebase. I don't think we * need * to offer support for multiple zkVMs from the start

emostov:
make sure to prefix you branch names with your names/initials

PR #341, opened 2024-11-06
feat: prefix some crates with `ivm-` and re-export helpful types

emostov:
The motivation for this is that we are using the crates in GTX and these updates make this easier. We are using this branch in https://github.com/GTX-Capital/gtx/pull/41/files

- re-export tonic types from protos so you don't need extra crates to send to CN
- prefix gtx used crate names with `ivm-`

follow ups:

- we should just prefix everything with `ivm-` for simplicity


crates/contracts/json/IJobManager.json
MaanavKhaitan, line 1: are these automatically generated? or did you have to copy these into here?

crates/sdk/test-utils/src/lib.rs
```diff
@@ -223,7 +223,7 @@ pub async fn create_and_sign_offchain_request(
     let job_params = JobParams {
         nonce,
         max_cycles,
-        // Need to use abi_encode_packed because the contract address
+        // Need to use ivm_abi_encode_packed because the contract address
```
MaanavKhaitan, line 226: I'm guessing this got changed by mistake lol
```suggestion
        // Need to use abi_encode_packed because the contract address
```

MaanavKhaitan (approved):

crates/contracts/json/IJobManager.json
emostov: copied in. My idea was to have a line in the make file that copies these in right after building the contracts

PR #342, opened 2024-11-06
feat: remove verifying key check

emostov:


crates/zkvm-executor/src/service.rs
```diff
@@ -78,6 +78,10 @@ where
 
     /// Checks the verifying key, executes a program on the given inputs, and returns signed output.
```
MaanavKhaitan, line 79: let's remove `Checks the verifying key` here

MaanavKhaitan (approved):
LGTM one nit

emostov:
i'm actually going to close this since getting through the merge conflicts is not worth it

PR #344, opened 2024-11-10
feat (coprocessor-node) comparing Risc0 vs SP1 zkVM performance benchmark

mayuridev:
closes #285. 

benchmarking SP1 vs Risc0 through various tests + generating graphs and visuals for docs. 

still WIP.


programs/sp1/mock-consumer/elf/riscv32im-succinct-zkvm-elf
emostov, line 1: we shouldn't commit this since its a build artifact so lets add the folder to git ignore like we do for `target`

emostov:
Overall looks great! I do think we should merge the "benchmarks", eventually it will just be sp1 once we rip out risc0

Cargo.toml
```diff
@@ -26,6 +26,8 @@ members = [
 	"programs/risc0/intensity-test/methods",
 	"programs/sp1/mock-consumer",
 	"programs/sp1/mock-consumer/program",
+	"programs/sp1/intensity-test/program",
+	"benches/vm_comparison",
```
emostov, line 30: For consistency, lets rename this directory to use kebab case: `vm-comparison`. Its just an arbitrary style thing to name crate directories kebab case

benches/vm_comparison/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "vm-comparison"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+
+[lints]
+workspace = true
+
+[[bin]]
```
emostov, line 10: Since you use `main.rs`, we by default build a binary called `vm-comparison`. So this is a little redundant. The bin key is useful when you have multiple bins or want the bin to have a different name then the crate

.github/workflows/rust.yml
```diff
@@ -122,6 +136,13 @@ jobs:
         uses: ./.github/actions/setup
         if: env.GIT_DIFF
 
+      - name: Build SP1 programs
+        if: env.GIT_DIFF
+        run: |
+          cd programs/sp1/mock-consumer/program && cargo prove build --elf-name mock-consumer-sp1-guest
```
emostov, line 142: so we can avoid this crate specific logic, can we structure the sp1 programs such that each program has an associated crate that has a build.rs file (so the same setup as what we do with risc0 now)? This crate can be included in the workspace, such that when you do `cargo build` it will build all the elfs. Heres an example build.rs file: https://github.com/succinctlabs/sp1/blob/dev/examples/io/script/build.rs. I think that it should be the same as using the CLI, but it makes sure we compile the program whenever we try to compile the crate containing the build file.

Once we migrate the example apps, we will have quite a few programs that need to be built and it will be nice to make sure all the elfs get built

programs/sp1/mock-consumer/src/lib.rs
```diff
@@ -1,2 +1,3 @@
 /// The ELF (executable and linkable format) file for the Succinct RISC-V zkVM.
-pub const MOCK_CONSUMER_SP1_GUEST_ELF: &[u8] = include_bytes!("../elf/riscv32im-succinct-zkvm-elf");
+pub const MOCK_CONSUMER_SP1_GUEST_ELF: &[u8] =
+    include_bytes!("../../../../target/elf-compilation/riscv32im-succinct-zkvm-elf/release/mock-consumer-sp1-guest");
```
emostov, line 3: Nice - I like having the elfs in the target dir. Is the `/elf-compilation` key something you added? I think its ok to skip that and have it all nested under `/riscv32im-succinct-zkvm-elf/` since that makes it clear its sp1 elf

emostov (approved):

PR #345, opened 2024-11-10
wip: ordered relaying queue & relayer refactor

emostov:
Changes
- Add db based Queue
- Add queues abstraction for looking up queue locks and all queue interaction
- Add relay strategy field to job
- Add logic to spin up background task to poll jobs that have been relay
- Do non blocking for unordered relays

TODO
- [ ]  Refactor all relay logic into a single service
    - [ ] Have all job writes in the service use a single writer task #353 
    - [ ] Potentially use parking lot mutex/rw lock on entire map to enforce only one table user at a time
    
- [ ] Move all DB writes to writer.
- [ ] https://crates.io/crates/num_cpus


crates/scripts/src/node_test.rs
```diff
@@ -224,6 +228,7 @@ async fn main() {
         request: encoded_job_request,
         signature,
         offchain_input: combined_offchain_input,
```
emostov, line 230: Need to make one of these ordered to check that the system works

emostov:
superseded #355

PR #352, opened 2024-11-12
fix: coprocessor-node dockerfile

emostov:
We broke the docker build when we renamed the coprocessor node crate to `ivm-coprocessor-node`. While fixing this I also realized we don't need the risc0 base image since we don't build any risc0 binaries, just the coprocessor-node binary. We also no longer rely on building the contracts at build time, the abi is committed in the crate.

Changes
- Don't install or use foundry
- Don't use a risc0 base image
- Add some apt installs

### DIFF
--- a/.github/workflows/docker-coproc.yml
+++ b/.github/workflows/docker-coproc.yml
@@ -15,7 +15,7 @@ env:
 jobs:
   docker-release:
     name: Docker release to Google Artifact Registry
-    runs-on: ubicloud-standard-16
+    runs-on: ubicloud-standard-30
 
     permissions:
       contents: 'read'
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-coproc.yml
```diff
@@ -2,6 +2,7 @@ name: docker-coproc
 
 on:
   workflow_dispatch:
+  pull_request:
```
MaanavKhaitan, line 5: why do we wanna build this every time we make a PR?

MaanavKhaitan (approved):
LGTM. were you able to test this?

.github/workflows/docker-coproc.yml
emostov: 1) so we don't silently break it and 2) so we can quickly test out PR branches

Issue #354, opened 2024-11-15
coprocessor-node: add table to track ID of non-relayed jobs.


emostov:
We should have a table to track non-relayed jobs. This is useful in case the node goes down and needs to find jobs that may have been executed but not relayed.

emostov:
This is a non-essential perf improvement we won't tackle soon

PR #355, opened 2024-11-15
feat: sync db writer, ordered relaying & relay coordinator

### LINKED ISSUE #300: coprocessor-node: add logic to enforce ordering for stateful job result relaying
In the case case of stateful offchain jobs, users will need their job results to hit in order. There is an edge case where job N+1 finishes before job N. We want to ensure that job N is relayed and accepted onchain before relaying job N+1. 

We also need to handle the case that one or more of these results is in the DLQ. Currently the DLQ is naive and just iterates the DB via cursor, which doesn't guarantee any specific ordering.

### LINKED ISSUE #353: coprocessor-node: enforce single DB writer
libmdbx will silently fail to write when there are too main concurrent attempts to write - this can clean to undefined states that are fatal to proper functioning.

```sh
"fields":{"message":"Process stalled, awaiting read-write transaction lock."},"target":"libmdbx"
```

 Their are two approaches I see: have a mutex to access DB writes, or have a single thread that does all db writes

emostov:
Primary features:
- Add relay queues for each consumer contract. We don't want a single queue because we don't want one users job blocking the relaying of all other jobs. Instead, we keep one queue per consumer. To achieve this we use a thread safe hashmap which tracks consumer_address => queue. We create the `Queues` abstraction which manages the queue access in an a way that efficiently uses locks
- Create a DB `Writer` that is a dedicated thread for all DB writes. All DB writes are sent to this thread and processed serially.
- Create a "relay coordinator" to manage all things relay related including adding job ids to queues, starting queue pollers, and actual relaying

Changes
- Add Queues abstraction for looking up queue locks and all queue interaction
- Add relay strategy field to job. This isn't signed over by the user so it can be manipulated. But if the user is using TLS straight to the node it should be ok.
- Rename job processor to job executor since it just executes jobs now and no longer does relaying
- Create RelayCoordinator
   - All queueing and job relaying is handled by the relay coordinator. This makes queuing logic simpler to reason about since it all happens in one place.
   - Spins up a queue poller anytime we get a queue request and see the queue is empty. The poller checks if the job is ready, and once ready it pops the associated id from the queue and relays it. The poller exits once the queue is empty.
- Create a DB writer - a loop that receives write requests over a channel and executes them. This operates in an OS thread since its a constant, long running process. All DB writes go through this process, ensuring that we don't encounter silent write failures when mdbx write locks timeout. This was necessary because we where getting a lot of failures when we tried spawning a tokio task per relay request. The upside is that we can write async in the background. The downside is that under load writes will get backed up in the channel. In the future we might be able to improve throughput by writing in batches.

Follow ups::
- Consider adding a table for seen job ids or job status to avoid writing the full job on intake. Also see if there are other ways we can avoid having to write the full job many times along the pipeline. https://github.com/InfinityVM/InfinityVM/issues/354
- Ordered dead letter queue https://github.com/InfinityVM/InfinityVM/issues/357
- Update coprocessor docs and diagram in book https://github.com/InfinityVM/InfinityVM/issues/363

Closes #353
Closes #300


crates/coprocessor-node/src/node.rs
```diff
@@ -87,7 +96,7 @@ pub struct NodeConfig<D> {
     /// Number of tx confirmations to wait for when submitting transactions.
     pub confirmations: u64,
     /// Job processor config values
-    pub job_proc_config: JobProcessorConfig,
+    pub job_proc_config: JobExecutorConfig,
```
emostov, line 99: need to remove this

crates/coprocessor-node/src/writer.rs
```diff
@@ -0,0 +1,105 @@
...
+    GenericRethMdbx(#[from] eyre::Report),
+    /// reth mdbx database backend error
+    #[error("mdbx (database): {0}")]
+    RethMdbx(#[from] reth_db::mdbx::Error),
+    /// reth database error
+    #[error("reth database: {0}")]
+    RethDbError(#[from] DatabaseError),
+}
+
+/// Table to write job too
```
emostov, line 35:
```suggestion
/// Write targets.
```

crates/coprocessor-node/src/relayer.rs
```diff
@@ -11,9 +24,25 @@ use alloy::{
...
+    sync::{mpsc::SyncSender, Arc},
+    time::Duration,
+};
+use tokio::{
+    sync::{mpsc::Receiver, oneshot},
+    time::interval,
+};
+use tracing::{error, info, span, Instrument, Level};
+
+/// Delay between retrying failed jobs, in milliseconds.m
```
MaanavKhaitan, line 44:
```suggestion
/// Delay between retrying failed jobs, in milliseconds.
```

crates/coprocessor-node/src/relayer.rs
```diff
@@ -205,7 +503,7 @@ impl JobRelayer {
         })?;
 
         let receipt =
-            pending_tx.with_required_confirmations(1).get_receipt().await.map_err(|error| {
+            pending_tx.with_required_confirmations(0).get_receipt().await.map_err(|error| {
```
MaanavKhaitan, line 506: why change from 1 to 0?

crates/coprocessor-node/src/relayer.rs
```diff
@@ -56,6 +85,275 @@ pub enum Error {
...
+                Some(job) => job,
+                // Edge case where the job has not been written yet
+                None => continue,
+            };
+
+            match job.status.status() {
+                JobStatusType::Done => {
+                    let _job_id = queues.pop_back(consumer).expect("queue is unexpected empty");
+                    debug_assert_eq!(job_id, _job_id);
+                    // relay the job below
```
MaanavKhaitan, line 213: let's remove this comment?

crates/coprocessor-node/src/relayer.rs
```diff
@@ -56,6 +85,275 @@ pub enum Error {
...
+        let queues = Queues::new();
+        let mut relay_rx = self.relay_rx;
+
+        while let Some(relay) = relay_rx.recv().await {
+            match relay {
+                Relay::Queue { consumer, job_id } => {
+                    let empty_queue = queues.peek_back(consumer).is_none();
+                    queues.push_front(consumer, job_id);
+
+                    // Queue pollers exit once a queue is empty. So we know that if the queue is
```
MaanavKhaitan, line 157: why do we need a separate queue poller for each consumer? can we have a single queue poller for all consumers and just keep that running on an OS thread?

crates/coprocessor-node/src/writer.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// Create a new instance of [`Self`]
+    pub fn new(db: Arc<D>, rx: mpsc::Receiver<WriterMsg>) -> Self {
+        Self { db, rx }
+    }
+
+    /// Start the job writer.
+    pub fn start_blocking(self) -> Result<(), Error> {
+        while let Ok((target, resp)) = self.rx.recv() {
+            let tx = self.db.tx_mut()?;
+            match target {
```
MaanavKhaitan, line 79: this is clean

crates/db/src/tables.rs
```diff
@@ -98,47 +98,51 @@ impl<'a> TryFrom<&'a Job> for JobParams<'a> {
...
 }
 
-/// Key to a table storing ELFs. The first byte of the key is the vm type
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, serde::Serialize, serde::Deserialize)]
-pub struct ElfKey(pub [u8; 32]);
+/// Key to a table storing ELFs. The first byte of the key is the vm type.
+#[derive(
+    Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, serde::Serialize, serde::Deserialize,
+)]
+pub struct Sha256Key(pub [u8; 32]);
```
MaanavKhaitan, line 126: why not name this `ElfKey` if it's used for the elf table?

crates/db/src/tables.rs
```diff
@@ -98,47 +98,51 @@ impl<'a> TryFrom<&'a Job> for JobParams<'a> {
 }
 
 /// Key to tables storing job metadata and failed jobs.
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, serde::Serialize, serde::Deserialize)]
-pub struct JobID(pub [u8; 32]);
+#[derive(
+    Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, serde::Serialize, serde::Deserialize,
+)]
+pub struct B256Key(pub [u8; 32]);
```
MaanavKhaitan, line 104: why not name this `JobID` or `JobKey` if it's used for the job table?

crates/scripts/src/node_test.rs
```diff
@@ -121,8 +121,12 @@ async fn main() {
     .await;
 
     // Submit a job to the coprocessor node
-    let submit_job_request =
-        SubmitJobRequest { request: encoded_job_request, signature, offchain_input: Vec::new() };
+    let submit_job_request = SubmitJobRequest {
+        request: encoded_job_request,
+        signature,
+        offchain_input: Vec::new(),
+        relay_strategy: RelayStrategy::Unordered as i32,
```
MaanavKhaitan, line 128: should we include this in the `JobRequest` so the user signs over it? I'm guessing this is something we want apps to decide so we should have them choose and sign over it?

crates/sdk/proto/src/descriptor.bin
MaanavKhaitan, line 1: was this binary included by mistake?

crates/scripts/src/node_test.rs
MaanavKhaitan: this feels like something we should just have in `JobRequest`

test/e2e/tests/mock_consumer.rs
```diff
@@ -86,22 +88,22 @@ async fn web2_job_submission_coprocessor_node_mock_consumer_e2e() {
...
             offchain_input: vec![],
+            relay_strategy: RelayStrategy::Unordered as i32,
         };
         let submit_job_response =
             args.coprocessor_node.submit_job(job_request).await.unwrap().into_inner();
         assert_eq!(submit_job_response.job_id, job_id);
 
         // wait for the job to be processed and relayed
-        tokio::time::sleep(tokio::time::Duration::from_secs(4)).await;
+        tokio::time::sleep(tokio::time::Duration::from_secs(3)).await;
```
MaanavKhaitan, line 98: any reason to decrease this? I think it was flaky on 3 seconds earlier

test/load-test/src/lib.rs
```diff
@@ -103,33 +103,44 @@ pub async fn get_offchain_request(
...
+    const RETRIES: u64 = MAX_RETRY_MS / RETRY_TIMEOUT_MS;
+
     let consumer_addr: Address =
         Address::parse_checksummed(consumer_addr(), None).expect("Valid address");
     let job_id = get_job_id(nonce, consumer_addr);
 
-    loop {
-        match get_result_status(user, job_id).await {
-            Ok(status) => {
-                // 2 is the status code for a completed job
```
MaanavKhaitan, line 114: we should add this comment back

test/load-test/src/main.rs
```diff
@@ -51,7 +51,7 @@ async fn main() -> Result<(), GooseError> {
         )
         .register_scenario(
             scenario!("LoadtestSubmitJob")
-                .set_wait_time(Duration::from_secs(1), Duration::from_secs(3))?
+                .set_wait_time(Duration::from_secs(0), Duration::from_secs(0))?
```
MaanavKhaitan, line 54: what's the point of having a wait time if it's 0? we can just remove the `set_wait_time`

test/load-test/src/main.rs
```diff
@@ -136,13 +137,19 @@ async fn loadtest_submit_job(user: &mut GooseUser) -> TransactionResult {
     let (encoded_job_request, signature) =
         get_offchain_request(nonce, &program_id, &encoded_intensity).await;
 
-    let submit_job_request =
-        SubmitJobRequest { request: encoded_job_request, signature, offchain_input: Vec::new() };
+    let submit_job_request = SubmitJobRequest {
+        request: encoded_job_request,
+        signature,
+        offchain_input: Vec::new(),
+        // If we want to just test max throughput, convert this to Unordered.
```
MaanavKhaitan, line 144: hmm seems like it's set to `Unordered` already

crates/coprocessor-node/src/relayer.rs
```diff
@@ -56,6 +85,275 @@ pub enum Error {
...
+        tokio::spawn(async move {
+            Self::start_job_retry_task(db, job_relayer, metrics, max_retries, writer_tx).await
+        });
+
+        let queues = Queues::new();
+        let mut relay_rx = self.relay_rx;
+
+        while let Some(relay) = relay_rx.recv().await {
+            match relay {
+                Relay::Queue { consumer, job_id } => {
```
MaanavKhaitan, line 153: i feel like it'd be clearer to name it `Relay::Ordered` and `Relay::Unordered`

MaanavKhaitan (commented):
🐐

Added a few comments

crates/coprocessor-node/src/relayer.rs
emostov: I need to investigate this, but the idea is that we want jobs to be able to be in strictly sequential blocks. So we want sequential jobs to in blocks 1,2,3 etc. My concern with requiring on conf is that we could only at beset do blocks 1,3,5 etc since we have. But I should probs just change this back

crates/zkvm-executor/src/service.rs
```diff
@@ -174,6 +174,7 @@ where
     pub async fn create_elf(&self, elf: &[u8], vm_type: VmType) -> Result<Vec<u8>, Error> {
         let vm = self.vm(vm_type)?;
 
+        // TODO: put this in a blocking task
```
emostov, line 171: TODO

crates/coprocessor-node/src/relayer.rs
emostov: The way this works is that for an `Ordered` request the intake will call `Relay::Queue`. After job execution, the JobExecutor will call `Relay::Now` regardless of type. This is because I wanted to keep as much code unaware of the relay ordering as possible - so now the JobExecutor can be unaware of relay ordering

crates/coprocessor-node/src/queue.rs
```diff
@@ -0,0 +1,168 @@
+//! Abstraction for global handle to all queues.
+
+use dashmap::DashMap;
+use parking_lot::Mutex;
+use std::{collections::VecDeque, sync::Arc};
+use tracing::instrument;
+
+type MutexQueue = Mutex<VecDeque<[u8; 32]>>;
+
+/// A thread safe in memory collection of job id queues, keyed by consumer contract address.
```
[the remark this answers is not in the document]
emostov: This is a collection of queues, not a single queue. If it was a single queue we could use a channel, but the issue is we need to track many queues

crates/coprocessor-node/src/relayer.rs
emostov: each consumer has a different queue in order to prevent one consumers jobs preventing another consumers from being relayed

crates/db/src/tables.rs
emostov: This is more reusable as we add more tables. This is an artifact for when I used the DB for queues. I decided to keep it

crates/scripts/src/node_test.rs
emostov: There is not any way to enforce the relay ordering, so I don't see the point of including in the part of the payload that gets posted on chain since its a waist. The one argument for signing it is that we don't want it modified in transit, but they presumably will have TLS to the CN

crates/sdk/proto/src/descriptor.bin
emostov: Its in main https://github.com/InfinityVM/InfinityVM/blob/main/crates/sdk/proto/src/descriptor.bin

test/e2e/tests/mock_consumer.rs
emostov: Artifact of me checking that the changes made things faster. I can revert it tho

crates/coprocessor-node/src/intake.rs
```diff
@@ -66,8 +85,10 @@ where
         exec_queue_sender: async_channel::Sender<Job>,
         zk_executor: ZkvmExecutorService<S>,
         max_da_per_job: usize,
+        writer_tx: SyncSender<WriterMsg>,
+        relay_tx: tokio::sync::mpsc::Sender<Relay>,
```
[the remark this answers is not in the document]
emostov: SyncSender has been update to crossbeam::Sender since you made this comment, but they are equivalent, crossbeam is just faster. The difference has to do with the type of code on the receiving side. The relay receiver is in a tokio task, so it is best to use an async. The writer receiver is synchronous code, so its best to use a blocking receiver

crates/db/src/tables.rs
emostov: https://github.com/InfinityVM/InfinityVM/pull/355#discussion_r1847470441

test/load-test/src/main.rs
emostov: done

crates/coprocessor-node/src/intake.rs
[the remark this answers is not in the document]
emostov: It had to do with if the context was sync or async. But I did some more research and it looks like flume is the best option since it supports bounded channels with both sync and async recv/send methods. Now everything uses flume

crates/coprocessor-node/src/cli.rs
```diff
@@ -242,7 +242,7 @@ impl Cli {
             http_eth_rpc: opts.http_eth_rpc,
             job_manager_address: opts.job_manager_address,
             confirmations: opts.confirmations,
-            job_proc_config: JobProcessorConfig {
+            job_proc_config: JobExecutorConfig {
```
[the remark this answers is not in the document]
emostov: renamed to RelayConfig

crates/coprocessor-node/src/queue.rs
```diff
@@ -0,0 +1,168 @@
...
+            Some(mutex) => mutex,
+        };
+        let queue = mutex.lock();
+
+        (*queue).back().copied()
+    }
+
+    /// Pop the element off back of the relay queue for `consumer_address`.
+    #[instrument(skip_all, level = "debug")]
+    pub fn pop_back(&self, consumer_address: [u8; 20]) -> Option<[u8; 32]> {
```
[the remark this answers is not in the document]
emostov: I rather keep these method names because they more closely reflect the methods we call on the underlying VecDeque https://doc.rust-lang.org/std/collections/struct.VecDeque.html

test/load-test/src/main.rs
emostov: Updated this so its configurable via a const

crates/coprocessor-node/src/relayer.rs
```diff
@@ -1,6 +1,19 @@
 //! Logic to broadcast job result onchain.
-
-use crate::metrics::Metrics;
+//!
+//! The primary type is the [`RelayCoordinator`]. The [`RelayCoordinator`] is in charge of managing
+//! broadcast queues on a per consumer basis, managing tasks queue polling tasks, and all final
+//! broadcasting logic. Other subsystems communicate with the [`RelayCoordinator`] by sending the
+//! [Relay] message over a mpsc channel. While the [Relay] channel its the only public API of this
```
emostov, line 6:
```suggestion
//! [Relay] message over a mpsc channel. While the [Relay] channel is the only public API for this
```

crates/coprocessor-node/src/queue.rs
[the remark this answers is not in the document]
MaanavKhaitan: hmm in VecDeque they push to the back and pop from the front

image

crates/coprocessor-node/src/queue.rs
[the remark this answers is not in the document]
MaanavKhaitan: ahh I see makes sense

crates/coprocessor-node/src/relayer.rs
MaanavKhaitan: ah gotcha

crates/db/src/tables.rs
MaanavKhaitan: cool cool

crates/coprocessor-node/src/relayer.rs
MaanavKhaitan: I understand that each consumer has a different queue, can we get away with having a single queue poller for all the consumers' queues though?

from `start_queue_poller` it seems like the queue poller doesn't wait on jobs being executed so it can't be blocked by a single consumer

crates/scripts/src/node_test.rs
MaanavKhaitan: Yep I was mainly suggesting this for simplicity + preventing someone from modifying it in transit. Down to go with whatever you think is best here

crates/coprocessor-node/src/relayer.rs
```diff
@@ -56,6 +89,278 @@ pub enum Error {
...
+            interval.tick().await;
+            let job = match get_job(db.clone(), job_id).expect("job get db error") {
+                Some(job) => job,
+                // Edge case where the job has not been written yet
+                None => continue,
+            };
+
+            match job.status.status() {
+                JobStatusType::Done => {
+                    let _job_id = queues.pop_back(consumer).expect("queue is unexpected empty");
```
MaanavKhaitan, line 217: what happens if the `relay_job_result` call after this fails for some reason? do we add the job back to the front of the queue so we try to relay this job again before relaying any subsequent jobs?

MaanavKhaitan: I wasn't able to find the code for that so guessing no. But if it's no, should we do this?

crates/coprocessor-node/src/relayer.rs
MaanavKhaitan: ah makes sense!

crates/coprocessor-node/src/server.rs
```diff
@@ -170,11 +170,15 @@ where
             return Err(Status::invalid_argument("program elf must not be empty"));
         }
 
-        let program_id = self
-            .intake_service
-            .submit_elf(req.program_elf, req.vm_type)
-            .await
-            .map_err(|e| Status::internal(format!("failed to submit ELF: {e}")))?;
+        // Deriving the program ID is expensive so we do it in a blocking task.
```
MaanavKhaitan, line 173: nice

crates/coprocessor-node/src/event.rs
```diff
@@ -149,9 +150,20 @@ where
                             blobs_sidecar: None,
                             relay_strategy: RelayStrategy::Unordered,
                         };
-                        if let Err(error) = self.intake.submit_job(job).await {
-                            error!(?error, ?event.jobID, "failed while submitting to job processor");
+
+                        for _ in 0..4 {
```
MaanavKhaitan, line 154: can we make the number of retries configurable / move this to a constant?

crates/coprocessor-node/src/relayer.rs
emostov: It goes into the failed jobs table if it fails. This isn't a perfect solution because if the job fails but the next does not then we have permanently messed up ordering. I'm not sure the ideal thing here, we could retry the same job for potentially hours even if its something thats just always going to fail so its a balance. I can change it to do that for now if we expect most issues to be temporary and resolve themselves. But if we expect most issues due to job results just not being able to be processed on chain then we need to solve this a different way: i.e. if job result submission always rollsback we need a way to submit results that doesn't trigger callbacks to ensure we can charge the user.

If all the relays end up failing then the failed jobs retry logic should respect ordering. It just retries at a much slower pace. I have this issue for improving the dead letter queue https://github.com/InfinityVM/InfinityVM/issues/357 to make sure its ordered, but considering that a non issue for now.

crates/coprocessor-node/src/relayer.rs
emostov: I updated this so now it respects the configured confirmations, which defaults to 1.

crates/scripts/src/node_test.rs
emostov: I added a note to use TLS here [6030969](https://github.com/InfinityVM/InfinityVM/pull/355/commits/60309692d3e1ac2ebcd6e09e1243c2c818b1a1da)

crates/coprocessor-node/src/relayer.rs
emostov: If we have 1000 consumer jobs ready to relay, we want to relay them in parallel not sequentially. Especially since each relay waits for a configured confirmation time - it could take 1000 blocks to relay 1000 jobs. 

One option I see is that we could have one poller that iterates through the the map and checks one at a time if a queue has a job ready to be relayed. If it does, then we can spawn a new task to do the relaying. The issue is that whenever we spawn a new task per relay it becomes complicated to know if the relay has completed. And for each consumer we only want to relay the next job once we are sure the current job has been fully included

crates/coprocessor-node/src/queue.rs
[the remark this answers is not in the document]
emostov: [0e61b1e](https://github.com/InfinityVM/InfinityVM/pull/355/commits/0e61b1ef710121a1b852c8866b61eb4ba48dc28a)

crates/coprocessor-node/src/relayer.rs
MaanavKhaitan: currently, if job with id `1` fails to relay but job with id `2` can be relayed, does the queue poller relay job `2` first or does it wait for job `1` to be retried?

emostov: It will retry job `1` 3 times and then put job one into one giant queue for all failed jobs. It will then try job `2`

emostov: > can you explain this more lol i dont understand
This relates to payments - we need to charge for jobs even if the consumer contract always rollsback when accepting the result

MaanavKhaitan: ah I think we can think about payments later when we're implementing it tbh. I think we'll be able to find a general solution that solves this then

I guess the main question now is: are we fine with retrying 3 times and then moving on to job 2? or do we wanna keep trying job 1 and not move on to job 2 until job 1 is relayed successfully?

MaanavKhaitan: > It will retry job 1 3 times and then put job one into one giant queue for all failed jobs. It will then try job 2

btw where is the code for making the queue poller wait for retrying job 1 3 times and only then trying job 2? I wasn't able to find this

emostov: @MaanavKhaitan my concern with indefinitely retrying is that we could be stuck on that job for infinity and have no way to unblock it

crates/coprocessor-node/src/queue.rs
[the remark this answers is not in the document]
MaanavKhaitan: thanku

crates/coprocessor-node/src/relayer.rs
MaanavKhaitan: i def agree, i think it's better to not indefinitely retry

crates/coprocessor-node/src/relayer.rs
MaanavKhaitan: ah makes sense, better to just have one queue poller per consumer then

MaanavKhaitan (approved):
LGTM assuming we already have this functionality:

> It will retry job 1 3 times and then put job one into one giant queue for all failed jobs. It will then try job 2

emostov:
> LGTM assuming we already have this functionality:
> 
> > It will retry job 1 3 times and then put job one into one giant queue for all failed jobs. It will then try job 2

Added in 4e8a6c7b62ffd8ca6f225b12934284f76b1370f8

PR #351, opened 2024-11-12
feat: add sp1 support & remove risc0 support

emostov:
TODO:
- [x] Ensure docker builds work
- [x] Migrate all programs to sp1
- [x] Remove all references to risc0

PR #356, opened 2024-11-15
feat: Port programs to SP1

mayuridev:
For #351


examples/matching-game/programs/src/lib.rs
```diff
@@ -1,85 +1 @@
-//! Binding to ZKVM programs.
-
-include!(concat!(env!("OUT_DIR"), "/methods.rs"));
-
-#[cfg(test)]
-mod tests {
```
MaanavKhaitan, line 6: again, why remove these tests?

programs/sp1/intensity-test/build.rs
```diff
@@ -0,0 +1,36 @@
+use sp1_build::{build_program_with_args, BuildArgs};
+use sp1_sdk::{HashableKey, ProverClient};
+use std::{fs, path::PathBuf};
+
+fn main() {
+    let output_dir = "elf-compilation/riscv32im-succinct-zkvm-elf/release/";
```
MaanavKhaitan, line 6: I feel like this build script is the same for clob + matching game + intensity test + mock consumer, what if we move this to some shared utils file and then each app just imports that and passes in some parameters like `elf_name` etc.?

programs/sp1/intensity-test/src/lib.rs
```diff
@@ -0,0 +1,22 @@
+use once_cell::sync::Lazy;
+use std::{fs, path::PathBuf};
+
+pub const INTENSITY_TEST_SP1_GUEST_ELF: &[u8] = include_bytes!(
+    "../../../../target/elf-compilation/riscv32im-succinct-zkvm-elf/release/intensity-test-sp1-guest"
+);
+
+static INTENSITY_TEST_ELF_ID: Lazy<[u8; 32]> = Lazy::new(|| {
+    let vkey_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
```
MaanavKhaitan, line 9: I feel like this logic within the `Lazy::new` is also very similar across clob + matching game + intensity test + mock consumer, so could be nice to move this into a shared utils file

test/e2e/tests/clob.rs
```diff
@@ -140,8 +135,9 @@ async fn state_job_submission_clob_consumer() {
         ];
         let clob_state3 = next_state(requests3.clone(), clob_state2.clone());
 
-        let mut interval = tokio::time::interval(tokio::time::Duration::from_secs(4));
+        let mut interval = tokio::time::interval(tokio::time::Duration::from_secs(10));
```
MaanavKhaitan, line 138: did we try with something like 8 and see if it's flaky? If it's not flaky then better to reduce the wait time

test/e2e/tests/clob.rs
```diff
@@ -178,13 +174,24 @@ async fn state_job_submission_clob_consumer() {
...
 
-            // Wait for the job to be processed
-            interval.tick().await;
+            let mut submit_job_response = None;
+            let mut retries = 3;
+            while retries > 0 {
+                match args.coprocessor_node.submit_job(job_request.clone()).await {
+                    Ok(response) => {
+                        submit_job_response = Some(response.into_inner());
+                        interval.tick().await;
```
MaanavKhaitan, line 184: i feel like we can move the `interval.tick().await` to after the while loop?

test/e2e/tests/clob.rs
```diff
@@ -178,13 +174,24 @@ async fn state_job_submission_clob_consumer() {
...
             let job_request =
                 SubmitJobRequest { request, signature, offchain_input: combined_offchain_input };
-            let submit_job_response =
-                args.coprocessor_node.submit_job(job_request).await.unwrap().into_inner();
 
-            // Wait for the job to be processed
-            interval.tick().await;
+            let mut submit_job_response = None;
+            let mut retries = 3;
+            while retries > 0 {
```
MaanavKhaitan, line 180: why do we need the retries? why was the test failing without the retries?

test/e2e/tests/clob.rs
```diff
@@ -227,12 +236,12 @@ async fn state_job_submission_clob_consumer() {
 #[tokio::test(flavor = "multi_thread")]
 async fn clob_node_e2e() {
     async fn test(mut args: Args) {
-        let mut interval = tokio::time::interval(tokio::time::Duration::from_secs(13));
+        let mut interval = tokio::time::interval(tokio::time::Duration::from_secs(30));
```
MaanavKhaitan, line 239: for all the wait time changes, just wanna ask if we've tested reducing these wait times and seeing how flaky the tests are

MaanavKhaitan: 30 seconds especially is quite long, but if it's actually flaky when the wait time is lower then I think it makes sense to keep this at 30

test/e2e/tests/matching_game.rs
```diff
@@ -237,8 +233,9 @@ async fn matching_game_server_e2e() {
             }
         );
 
-        // Give the batcher some time to process.
+        // Give the batcher some time to process
```
MaanavKhaitan, line 236:
```suggestion
        // Give the batcher some time to process.
```

test/e2e/tests/clob.rs
MaanavKhaitan: same thing goes for all the sleeps in this file, `matching_game.rs`, etc.

test/load-test/src/main.rs
```diff
@@ -91,7 +92,7 @@ pub async fn submit_first_job() -> Result<(), Box<dyn std::error::Error>> {
             Address::parse_checksummed(consumer_addr(), None).expect("Valid address");
         let (encoded_job_request, signature) = get_offchain_request(
             nonce,
-            &INTENSITY_TEST_GUEST_ID.iter().flat_map(|&x| x.to_le_bytes()).collect::<Vec<u8>>(),
+            &get_mock_consumer_sp1_guest_elf_id(),
```
MaanavKhaitan, line 95: hmm we should use the intensity test elf here?

MaanavKhaitan (commented):
This is great! added a few comments

.github/workflows/rust.yml
```diff
@@ -71,7 +71,7 @@ jobs:
         uses: ./.github/actions/setup
         if: env.GIT_DIFF
 
-      - run: make test-all
+      - run: make BUILD=release test-all
```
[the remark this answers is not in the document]
emostov: If so, why is it necessary and should we add it to the cargo command

examples/clob/programs/sp1/build.rs
```diff
@@ -0,0 +1,33 @@
...
+    let elf_path = PathBuf::from("program/").join(output_dir).join(elf_name);
+
+    println!("cargo:rerun-if-changed={}", elf_path.display());
+
+    if let Ok(program_elf) = fs::read(&elf_path) {
+        let (_, vk) = ProverClient::new().setup(&program_elf);
+        let vk_bytes = vk.hash_bytes().to_vec();
+        let vk_path = elf_path.with_extension("vkey");
+
+        bincode::serialize_into(
```
emostov, line 25: Looking at this, can we just write the raw bytes instead of serializing them with bincode? I know I mentioned using bincode, but it doesn't seem neccesary if we can just write the `vk_bytes` as is directly

examples/clob/programs/sp1/src/lib.rs
```diff
@@ -0,0 +1,16 @@
+use once_cell::sync::Lazy;
+use std::fs;
+
+pub const CLOB_ELF: &[u8] = include_bytes!("../program/elf/clob-sp1-guest");
+
```
emostov, line 5: If we ditch the bincode, I think we could just do this, which would be awesome because we can then use this public const everywhere

```suggestion
pub const CLOB_ELF_ID: &[u8] = include_bytes!("../program/elf/clob-sp1-guest.vkey")
```

examples/clob/programs/sp1/build.rs
```diff
@@ -0,0 +1,33 @@
...
+    };
+    build_program_with_args("program", args);
+
+    let elf_path = PathBuf::from("program/").join(output_dir).join(elf_name);
+
+    println!("cargo:rerun-if-changed={}", elf_path.display());
+
+    if let Ok(program_elf) = fs::read(&elf_path) {
+        let (_, vk) = ProverClient::new().setup(&program_elf);
+        let vk_bytes = vk.hash_bytes().to_vec();
```
emostov, line 22: why the hash? How big is this if we don't hash it

examples/clob/programs/sp1/src/lib.rs
emostov: If this makes sense, we should use the same pattern for all the programs

examples/clob/programs/sp1/build.rs
mayuridev: just thought that we could use fixed size so we dont have super big elf files for bigger programs, they might get p big depending on how complex they r

MaanavKhaitan: now after I merged in #364 this is not an issue anymore

examples/matching-game/programs/sp1/build.rs
```diff
@@ -0,0 +1,33 @@
...
+        let vk_bytes = vk.hash_bytes().to_vec();
+        let vk_path = elf_path.with_extension("vkey");
+
+        bincode::serialize_into(
+            fs::File::create(&vk_path).expect("Failed to create vkey file"),
+            &vk_bytes,
+        )
+        .expect("Failed to serialize verifying key");
+
+        println!("cargo:rerun-if-changed={}", vk_path.display());
```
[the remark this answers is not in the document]
MaanavKhaitan: seems like @mayuridev updated this now

crates/sdk/sp1-utils/src/lib.rs
```diff
@@ -4,7 +4,10 @@ use sp1_build::{build_program_with_args, BuildArgs};
 use std::{fs, path::PathBuf};
 
 pub fn build_sp1_program(elf_name: &str, program_dir: &str, output_dir: &str) {
-    // We need to get the workspace root to convert all paths to absolute paths
+    // Track the Cargo.toml and src directory of the program
+    println!("cargo:rerun-if-changed={}/Cargo.toml", program_dir);
```
MaanavKhaitan, line 8: I just ran `cargo build` without changing anything in the program files but it's still building `mock_consumer` and `intensity-test` (for matching-game and clob it seems to use the previous builds). Is there something we can do to fix this?

examples/clob/programs/src/lib.rs
```diff
@@ -1,16 +1,7 @@
-use ivm_sp1_utils::get_program_id;
-use once_cell::sync::Lazy;
-
 pub const CLOB_ELF: &[u8] = include_bytes!("../../../../target/sp1/clob/clob-sp1-guest");
 
-static CLOB_PROGRAM_ID: Lazy<[u8; 32]> = Lazy::new(|| {
-    let program_id_path = "target/sp1/clob/clob-sp1-guest.vkey";
```
MaanavKhaitan: why remove this? now someone building an app needs to make sure to update the relative path in `CLOB_PROGRAM_ID = *include_bytes!(...` which I don't think is ideal

MaanavKhaitan: actually nvm I think this is fine

mayuridev: Yeah tbh I made this change before your commit fixing the relative sp1 path issues so it's not exactly optimal but I had it added after the comments on the last commits in this PR

MaanavKhaitan: can we add it back

MaanavKhaitan: we also use `ivm_sp1_utils::get_program_id` in the foundry template so we can avoid dealing with relative paths there. So if we can add these things back that'd be great

crates/sdk/sp1-utils/src/lib.rs
```diff
@@ -4,26 +4,26 @@ use sp1_build::{build_program_with_args, BuildArgs};
...
     let elf_path_abs = output_dir_abs.join(elf_name);
-    if let Ok(program_elf) = fs::read(&elf_path_abs) {
-        let program_id = ivm_zkvm::Sp1.derive_program_id(&program_elf).unwrap();
-        let program_id_path_abs = elf_path_abs.with_extension("vkey");
+    let program_id_path_abs = elf_path_abs.with_extension("vkey");
+
+    println!("cargo:rerun-if-changed={}/Cargo.toml", program_dir);
+    println!("cargo:rerun-if-changed={}/src", program_dir);
+
+    if !elf_path_abs.exists() || !program_id_path_abs.exists() {
```
MaanavKhaitan, line 16: hmm so if the elf file and program ID file both exist, then this will not build the ELF again? I think that logic is wrong

MaanavKhaitan (dismissed):

test/e2e/tests/clob.rs
```diff
@@ -140,8 +135,9 @@ async fn state_job_submission_clob_consumer() {
         ];
         let clob_state3 = next_state(requests3.clone(), clob_state2.clone());
 
-        let mut interval = tokio::time::interval(tokio::time::Duration::from_secs(4));
+        let mut interval = tokio::time::interval(tokio::time::Duration::from_secs(8));
         interval.tick().await; // First tick processes immediately
+        interval.tick().await;
```
MaanavKhaitan, line 140: why do we have two ticks here?

test/e2e/tests/clob.rs
```diff
@@ -398,6 +394,7 @@ async fn clob_node_e2e() {
         interval.tick().await;
 
         // Check that balances have been updated on chain from the batch.
+        sleep(Duration::from_secs(15)).await;
```
MaanavKhaitan, line 397: btw did we try and see if we can use a shorter sleep duration for this? if yes then it's fine

if no then maybe we should try reducing it and check if that works

Cargo.toml
```diff
@@ -180,9 +174,9 @@ serde_json = { version = "1.0", default-features = false }
 sha2 = { version = "0.10", features = ["std"], default-features = false }
 strum = { version = "0.26.3", features = ["derive"], default-features = false}
 
-sp1-zkvm = { git = "https://github.com/succinctlabs/sp1.git", rev = "8a62caa0b8893aabe8e9477d4568c20387ae4896", features = ["libm", "lib"], default-features = false }
-sp1-sdk = { git = "https://github.com/succinctlabs/sp1.git", rev = "8a62caa0b8893aabe8e9477d4568c20387ae4896", default-features = false }
-sp1-helper = { git = "https://github.com/succinctlabs/sp1.git", rev = "8a62caa0b8893aabe8e9477d4568c20387ae4896", default-features = false }
+sp1-zkvm = { git = "https://github.com/succinctlabs/sp1.git", rev = "37629f6", features = ["libm", "lib"], default-features = false }
+sp1-sdk = { git = "https://github.com/succinctlabs/sp1.git", rev = "37629f6", features = ["network"], default-features = false }
```
emostov, line 178: any idea why we need this network feature?

crates/sdk/sp1-utils/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+
+pub fn build_sp1_program(elf_name: &str, program_dir: &str, output_dir: &str) {
+    let metadata = MetadataCommand::new().no_deps().exec().unwrap();
+    let workspace_root = metadata.workspace_root;
+    let output_dir_abs = PathBuf::from(workspace_root).join(output_dir);
+    let elf_path_abs = output_dir_abs.join(elf_name);
+    let program_id_path_abs = elf_path_abs.with_extension("vkey");
+
+    println!("cargo:rerun-if-changed={}/Cargo.toml", program_dir);
+    println!("cargo:rerun-if-changed={}/src", program_dir);
```
emostov, line 14: Maybe we should remove these for now and add back some variation when we fix builds? I don't like that they are hidden in the helper function and rather have them explicit in build.rs

test/e2e/tests/clob.rs
emostov: We discussed in messaging - will revisit in follow up

emostov (approved):
SHIPPP

Cargo.toml
MaanavKhaitan: doesn't compile without it

crates/sdk/sp1-utils/src/lib.rs
MaanavKhaitan: done

test/e2e/tests/clob.rs
mayuridev: Yeah I tried shorter but they kept on giving me timing errors and it took like 30 mins per test run lol

PR #364, opened 2024-11-19
refactor: Rename to `program_id` and reorganize program folders

MaanavKhaitan:
# What

- Renames "verification key" and "elf ID" to "program ID". We were using all these three names to refer to the same thing and it was getting confusing and hard to search for things
- Renames `MOCK_CONSUMER_SP1_GUEST_ELF` to `MOCK_CONSUMER_ELF` etc.
- Each program's folder previously had an `sp1` folder within it which then contained the actual program. Removes this intermediary `sp1` folder since for now we are only supporting sp1 and want to simplify the repo structure
- Earlier, the ELF files for `intensity-test` and `mock-consumer` were being generated in an `elf-compilation` folder at the root of the repo, while the ELF files for `matching-game` and `clob` were generated in an `program/elf` folder within each of these app's folders. Updates this so that the ELF files are generated within `program/elf` for all programs instead of the `elf-compilation` folder

# Testing

`make test-all`: Passes


programs/sp1/mock-consumer/build.rs
```diff
@@ -4,6 +4,6 @@ fn main() {
     build_sp1_program(
         "mock-consumer-sp1-guest",
         "program",
-        "elf-compilation/riscv32im-succinct-zkvm-elf/release/",
+        "programs/sp1/mock-consumer/program/elf/",
```
MaanavKhaitan, line 7: for some reason we need to provide a path relative to the root of the repo here

MaanavKhaitan: i'm guessing it's some internal sp1 thing

emostov (commented):
I like the goal of consistency. My only gripe is that we should have all build artifacts under `target/` to follow rust workspace convention. So this would mean having both the program ID and elf inside some target subdirectory

crates/sp1-utils/src/lib.rs
```diff
@@ -10,28 +10,38 @@ pub fn build_sp1_program(elf_name: &str, program_dir: &str, output_dir: &str) {
     };
     build_program_with_args(program_dir, args);
 
-    let elf_path = PathBuf::from(program_dir).join(output_dir).join(elf_name);
+    // If output_dir contains program_dir, extract the substring starting from program_dir.
+    // Otherwise, join program_dir with output_dir. This is to handle cases where
+    // the output_dir is provided relative to the root of the repo, while program_dir
+    // is provided relative to the current directory.
+    let elf_path = if output_dir.contains(program_dir) {
```
emostov, line 17: I think we can hardcode the output dir to `repo-root/target/sp1/{program-name}`

MaanavKhaitan: updated

crates/sp1-utils/src/lib.rs
```diff
@@ -10,28 +10,46 @@ pub fn build_sp1_program(elf_name: &str, program_dir: &str, output_dir: &str) {
...
+
+    // Construct elf path relative to current dir with correct number of parent dirs.
+    let elf_path_string =
+        format!("{}/{}", "../".repeat(dir_count), output_dir.trim_start_matches("../"));
+    let elf_path = PathBuf::from(elf_path_string).join(elf_name);
 
     println!("cargo:rerun-if-changed={}", elf_path.display());
 
+    println!("cargo:warning=WOOOOO elf_path: {}", elf_path.display());
+    println!("cargo:warning=WOOOOO current directory: {}", std::env::current_dir().unwrap().display());
```
emostov, line 28: do we want these as cargo warnings?

crates/sp1-utils/src/lib.rs
```diff
@@ -10,28 +10,46 @@ pub fn build_sp1_program(elf_name: &str, program_dir: &str, output_dir: &str) {
...
+    let infinity_vm_pos = current_dir.rfind("InfinityVM").unwrap_or(0);
+    let path_after_infinity = &current_dir[infinity_vm_pos..];
+    let dir_count = path_after_infinity.matches('/').count();
+
+    // Construct elf path relative to current dir with correct number of parent dirs.
+    let elf_path_string =
+        format!("{}/{}", "../".repeat(dir_count), output_dir.trim_start_matches("../"));
+    let elf_path = PathBuf::from(elf_path_string).join(elf_name);
 
     println!("cargo:rerun-if-changed={}", elf_path.display());
```
emostov, line 25: maybe out of scope, but i think this re-run logic doesn't make sense - we don't want to rebuild if the artifacts changes - we want to rebuild if the source changes

examples/clob/programs/src/lib.rs
```diff
@@ -1,4 +1,16 @@
...
+use once_cell::sync::Lazy;
+
+pub const CLOB_ELF: &[u8] = include_bytes!("../../../../target/sp1/clob/clob-sp1-guest");
+
+static CLOB_PROGRAM_ID: Lazy<[u8; 32]> = Lazy::new(|| {
+    let program_id_path = "target/sp1/clob/clob-sp1-guest.vkey";
+    get_program_id(program_id_path)
+});
+
+pub fn get_clob_program_id() -> [u8; 32] {
```
emostov, line 11: do we need this fn if we have the static? I haven't used statics before so not sure

MaanavKhaitan: I like having this function because then consumers of the program ID don't need to worry about the program ID being `Lazy<>` and needing to dereference it etc.

crates/sdk/sp1-utils/Cargo.toml
```diff
@@ -1,9 +1,12 @@
...
 
 [dependencies]
 bincode = { workspace = true }
 sp1-build = { workspace = true }
-ivm-zkvm = { workspace = true }
\ No newline at end of file
+ivm-zkvm = { workspace = true }
+cargo_metadata = { version = "0.18" }
+
+[build-dependencies]
```
emostov, line 12: i think we can remove this lline

MaanavKhaitan: ah thanks just fixed

PR #373, opened 2024-11-24
feat: Use sp1 executor directly

MaanavKhaitan:


MaanavKhaitan:
Closing for now because sp1 said this is not possible rn. They might expose the "execute" functionality in the sp1-sdk and we can implement this when they do that

PR #374, opened 2024-11-24
fix: Move intensity-test and mock-consumer programs to their own workspace

MaanavKhaitan:
# What

Earlier, the `cargo rerun` logic wasn't working for the `intensity-test` and `mock-consumer` programs, but was working for matching game and clob.

- Moves these programs to their own workspace. This was the one difference between these programs and the matching game/clob. So far, this seems to be working

# Testing

Ran `cargo build` a few times and made changes to some other files, and verified that we're not rebuilding the ELFs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,9 +24,7 @@ members = [
 	"test/e2e",
 	"test/load-test",
 	"programs/mock-consumer",
-	"programs/mock-consumer/program",
 	"programs/intensity-test",
-	"programs/intensity-test/program",
 ]
 
 # Explicitly set the resolver to version 2, which is the default for packages with edition >= 2021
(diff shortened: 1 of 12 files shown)


emostov:
It doesn't rebuild because `cargo build` just builds the workspace. But on a fresh clone/ while iterating on the programs we want to rebuild them. CI does fresh clones and its passing here so I am probably missing something. I am down to try this out; lets just keep an eye on it

emostov (approved):

Cargo.toml
```diff
@@ -24,9 +24,7 @@ members = [
 	"test/e2e",
 	"test/load-test",
 	"programs/mock-consumer",
-	"programs/mock-consumer/program",
 	"programs/intensity-test",
-	"programs/intensity-test/program",
 ]
```
emostov, line 29: theres a chance we might need the exclude key https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields. But if it isn't complaining I guess we don't need it?

MaanavKhaitan: ah yes you're right. This solved another issue so we don't have two builds for each ELF now

MaanavKhaitan: just added this

emostov: oh interesting

PR #376, opened 2024-11-25
chore: remove unused sp1 build deps

emostov:
Changes
- Guest program had build deps event tho they didn't have build logic. We remove these
- Reduce timeouts in e2e tests and force tokio test worker threads in some places to reduce chances of slow tasks blocking. Since we refactored the CN to use standard OS threads, job execution should be faster and we should be able to reduce tiemouts

### DIFF
--- a/test/e2e/tests/clob.rs
+++ b/test/e2e/tests/clob.rs
@@ -24,7 +24,7 @@ use ivm_proto::{GetResultRequest, RelayStrategy, SubmitJobRequest, SubmitProgram
 use tokio::time::{sleep, Duration};
 
 #[ignore]
-#[tokio::test(flavor = "multi_thread")]
+#[tokio::test(flavor = "multi_thread", worker_threads = 8)]
 async fn state_job_submission_clob_consumer() {
     async fn test(mut args: Args) {
         let anvil = args.anvil;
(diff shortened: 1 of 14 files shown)


test/e2e/tests/clob.rs
```diff
@@ -135,9 +135,6 @@ async fn state_job_submission_clob_consumer() {
         ];
         let clob_state3 = next_state(requests3.clone(), clob_state2.clone());
 
-        let mut interval = tokio::time::interval(tokio::time::Duration::from_secs(8));
```
MaanavKhaitan, line 138: just wanna make sure, removing this won't make things more flaky on CI?

test/e2e/tests/clob.rs
```diff
@@ -182,7 +179,7 @@ async fn state_job_submission_clob_consumer() {
                 args.coprocessor_node.submit_job(job_request).await.unwrap().into_inner();
 
             // Wait for the job to be processed
-            interval.tick().await;
+            sleep(Duration::from_secs(8)).await;
```
MaanavKhaitan, line 182: should we reduce this to 8 in the `mock-consumer.rs` e2e test too?

MaanavKhaitan (approved):
mainly just asking if this will make CI more flaky

image

test/e2e/tests/clob.rs
emostov: if it does we can just ask them

PR #377, opened 2024-11-25
feat!: skip verifying key derive in tests

emostov:
Changes
- breaking: add program ID to submit program request. This allows us to pass this back to the user when we want to skip program ID derives
- add option to opt out of checking program IDs. This isn't available in the CLI and we instead just configure it when setting up testing
- regen protos. Looks like we are on a new proto dep version so generated code is a little different


crates/coprocessor-node/src/event.rs
```diff
@@ -163,11 +163,6 @@ where
                             let backoff = i * SUBMIT_JOB_BACKOFF_BASE_MS;
                             sleep(Duration::from_millis(backoff as u64)).await;
                         }
-
-                        error!(
```
emostov, line 167: this error was incorrect - it prints any time we successfully send a job requests

crates/scripts/src/local.rs
```diff
@@ -132,21 +133,26 @@ async fn main() {
         let mut coproc_client =
             CoprocessorNodeClient::connect(client_coproc_grpc.clone()).await.unwrap();
         info!("Submitting CLOB ELF to coprocessor node");
-        let submit_program_request =
-            SubmitProgramRequest { program_elf: CLOB_ELF.to_vec(), vm_type: VmType::Sp1.into() };
+        let submit_program_request = SubmitProgramRequest {
+            program_elf: CLOB_ELF.to_vec(),
+            vm_type: VmType::Sp1.into(),
+            program_id: get_matching_game_program_id().to_vec(),
```
MaanavKhaitan, line 139: this should be `get_clob_program_id()`

crates/sdk/proto/src/descriptor.bin
MaanavKhaitan, line 1: should this be committed to git?

MaanavKhaitan (approved):
LGTM, a few nits

crates/sdk/proto/src/descriptor.bin
emostov: ya we have been doing it so far

crates/scripts/src/local.rs
emostov: nice ty

PR #382, opened 2024-11-26
fix: Update docs and use `dotenvy` instead of `dotenv`

### LINKED ISSUE #380: dotenv is unmaintained, we should use dotenvy.
We should migrate to dotenvy

MaanavKhaitan:
# What

Based on Avi's feedback:

- Uses `dotenvy` instead of `dotenv` closes #380 
- Points to `CONTRIBUTING.md` from the book
- Adds brief info about public/private inputs and shortening the fraud proof window by submitting a ZK validity proof

# Testing

`just test-all`: Passes


book/src/integration/using-program.md
```diff
@@ -26,3 +26,6 @@ Offchain job requests can be sent by a user, your app, or any authorized third-p
 
 You can read more in the [Offchain Jobs](./offchain.md) section.
 
+### Public and private inputs
+
+For now, all inputs are public. We are working on adding support for private inputs.
```
emostov, line 31: nit: we aren't working on it per se but maybe  `We plan on adding support for private inputs in the future.`

emostov (approved):
LGTM - one non essential nit

book/src/integration/using-program.md
lucasege: nit: it seems like adding support for private inputs would be incompatible with the current IVM design. so yeah i agree "plannign to eventually maybe add support" is probably correct

MaanavKhaitan: updated

Avi-D-coder (approved):

lucasege (approved):

PR #386, opened 2024-11-28
fix: reducing timeouts in tests

### LINKED ISSUE #368: sp1: Investigate reducing timeouts in tests
Currently, we have timeouts/sleeps for 15-20 seconds and multiple consecutive interval ticks in the e2e tests. We want to look into whether we can reduce these timeouts and ticks

mayuridev:
closes #368.


MaanavKhaitan (approved):
LGTM.  As long as we're fairly confident that this doesn't make the tests more flaky

PR #387, opened 2024-11-30
refactor: use program ID path instead of lazy

MaanavKhaitan:
# What

- Uses the program ID path instead of lazy init

# Testing

`make test-all`: Passes


Avi-D-coder (approved):
lgtm

PR #392, opened 2024-12-06
wip: ivm-exec

### LINKED ISSUE #384: custom reth node for sub second block times
Now that https://github.com/paradigmxyz/reth/pull/12900 is merged, it should be easier to create a reth based node that does not check the block timestamp and can do subsecond block times. Right now we have a fork for this, but we want to transition to uses the reth node builder.

emostov:
closes #384 

TODO: 
- [ ] check that engine validator timestamp checked is definitely getting threaded down to new engine


emostov:
superseded by #394

PR #394, opened 2024-12-11
feat: introduce ivm-exec

emostov:
closes https://github.com/InfinityVM/InfinityVM/issues/384

Changes
- add `ivm-exec` binary, infinity's reth based execution client
- define custom engine validator so we can avoid timestamp check
- bump reth and alloy


bin/exec/src/engine.rs
```diff
@@ -0,0 +1,102 @@
...
+    ) -> Result<(), EngineObjectValidationError> {
+        validate_version_specific_fields(self.chain_spec(), version, payload_or_attrs)
+    }
+
+    fn ensure_well_formed_attributes(
+        &self,
+        version: EngineApiMessageVersion,
+        attributes: &Types::PayloadAttributes,
+    ) -> Result<(), EngineObjectValidationError> {
+        validate_version_specific_fields(self.chain_spec(), version, attributes.into())
```
lucasege, line 71: is this meant to call into reth's default `ensure_well_formed_attributes`?

crates/db/src/tables.rs
```diff
@@ -6,8 +6,8 @@ use ivm_abi::JobParams;
 use ivm_eip4844::BlobTransactionSidecar;
 use ivm_proto::{JobStatus, RelayStrategy};
 use reth_db::{
-    table::{Decode, Encode},
-    tables, DatabaseError, TableType, TableViewer,
+    table::{Decode, Encode, TableInfo},
```
lucasege, line 9: are these imports still necessary? just surprised to see an import added but no code using it

examples/clob/node/src/db/tables.rs
```diff
@@ -1,7 +1,7 @@
 //! Tables for the database.
 
 use crate::db::models::{ClobStateModel, RequestModel, ResponseModel, VecDiffModel};
-use reth_db::{tables, TableType, TableViewer};
+use reth_db::{table::TableInfo, tables, TableSet, TableType, TableViewer};
```
lucasege, line 4: same question

lucasege (approved):
Looks good 1 question

crates/db/src/tables.rs
emostov: I didn't look too deeply, but it appears these traits are now expected after bumping reth

emostov: The traits where added in this PR https://github.com/paradigmxyz/reth/pull/13130/files. I think they should ideally be added to the tables macro so we don't need to import them, but doesn't seem worth the effort to upstream

bin/exec/src/engine.rs
emostov: I think I need to bring in https://reth.rs/docs/reth_ethereum_engine_primitives/struct.EthereumEngineValidator.html as another member of our validator. Let me look into this

emostov: I updated to pull in the `EthereumEngineValidator`

PR #395, opened 2024-12-11
feat: add custom transaction pool for ivm-exec

emostov:
Changes
- Add ivm config toml file. Currently it allows configuring allow lists for transaction sender, and transaction to field. 
- Add CLI flag for specifying a path to the config file
- Add custom transaction pool with custom validation logic based on allow lists
- Add ivm-exec readme


bin/exec/README.md
```diff
@@ -0,0 +1,30 @@
+# Infinity Execution Client
+
+`ivm-exec` is the Infinity execution client. It is built using reth and closely follows ethereum specs except in two edge cases:
+
+1) It ignores timestamp checks in the Engine API. Timestamps are denominated in seconds and the Engine API dictates that the timestamp for an execution payload being requested must be greater then the previous execution payload. Ignoring checks allows us to request multiple execution payloads in the same second and thus achieve sub second block times.
```
lucasege, line 5: If I was a reader my #1 question would be "at what cost", e.g. what's the tradeoff. presumably the Engine API added this check for a reason, is there any downside to us ignoring it?
If we don't know, that's fine, but that just comes to my mind

bin/exec/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+The toml config file should look like 
+
+```toml
+[transaction_allow]
+# Allow all transactions regardless of allow lists
+all = false
+# Allow transactions with to field that contain one of these addresses
+to = ["0x1111111111111111111111111111111111111111"]
+# Allow transactions signed by one of these addresses
```
lucasege, line 28: nit: can we specify the relationship between `to` and `sender`? AFAIK its an OR (e.g. a tx doesn't need to match both conditions), but we should specify

lucasege: it's kinda implied in your comments but i don't think it would hurt to be verbose here since i anticipate us modifying this config under stress

bin/exec/src/pool/validator.rs
```diff
@@ -0,0 +1,205 @@
...
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct IvmTransactionAllowConfig {
+    all: bool,
+    to: HashSet<Address>,
+    sender: HashSet<Address>,
+}
+
+impl Default for IvmTransactionAllowConfig {
+    fn default() -> Self {
+        Self { sender: HashSet::default(), to: HashSet::default(), all: true }
```
lucasege, line 27: I'm a little bit nervous about making `all: true` the default. I'll raise this in slack, but idk if the general consensus for infra is to make the default case optimistic or pessimistic. e.g. if all of our nodes die and all of our ivmconfig files get deleted do we want to to default permissive or not

bin/exec/README.md
emostov: 9b47ad77398e3d471952ff43580076546e999d0d

bin/exec/src/pool/validator.rs
emostov: I made it default to false [6ebf297](https://github.com/InfinityVM/InfinityVM/pull/395/commits/6ebf297ad34c7bba6592fbff3df542982ae7325c)

bin/exec/README.md
emostov: I linked the spec https://github.com/InfinityVM/InfinityVM/commit/9b47ad77398e3d471952ff43580076546e999d0d. I think its just to make sure the block is valid and not going "backwards in time"

emostov: One thing is we could just change the check to be `>=` instead of just `>`

bin/exec/src/pool/validator.rs
```diff
@@ -0,0 +1,205 @@
...
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::IvmTransactionAllowConfig;
+    use alloy::primitives::Address;
+    use std::collections::HashSet;
+
+    // Helpers to set/get values for tests
```
MaanavKhaitan, line 169: do we need to add actual tests here or is it just helpers?

bin/exec/README.md
MaanavKhaitan: hmm I thought that's what we're doing lol (changing the check to be `>=`). If not this, what are we doing?

MaanavKhaitan (approved):
LGTM

bin/exec/src/pool/validator.rs
emostov: I just put helpers here since we have access to the private fields in this scope. The validator is tested in the pool tests

PR #396, opened 2024-12-16
chore: dockerfile for `ivm-exec`

emostov:


Dockerfile.exec
```diff
@@ -0,0 +1,33 @@
...
+FROM debian:bookworm-slim
+WORKDIR /app
+RUN apt-get update && \
+  apt-get install -y --no-install-recommends ca-certificates && \
+  rm -rf /var/lib/apt/lists/*
+
+COPY --from=builder /app/target/release/ivm-exec /usr/local/bin/ivm-exec
+COPY --from=grafana/promtail:3.0.0 /usr/bin/promtail /usr/local/bin
+
+EXPOSE 30303 30303/udp 9001 8545 8546
```
emostov, line 32: I just pulled these from the reth dockerfile https://github.com/paradigmxyz/reth/blob/82f21a3e39d4e4ce115b11fc6af849c783271363/Dockerfile#L51

Dockerfile.exec
```diff
@@ -0,0 +1,33 @@
...
+ENTRYPOINT ["/usr/local/bin/ivm-exec"]
```
emostov, line 33: Not sure if we want this, we don't have an entry point in the coproc

twoshark: We override it so it doesn't really matter

twoshark (approved):

PR #399, opened 2024-12-17
fix: docker binary name for ivm-exec

twoshark:


emostov (approved):
I think we want to change the entrypoint as well to be `/usr/local/bin/reth`

PR #409, opened 2024-12-21
exec: zero gas fee execution

emostov:
This PR adds a custom EVM config for reth that overrides some `revm` handler functions to ignore charging and refunding gas to accounts.


bin/exec/Cargo.toml
```diff
@@ -22,6 +26,16 @@ serde = { workspace = true }
 
 [dev-dependencies]
 reth-provider = { workspace = true, features = ["test-utils"] }
-# This is a bit hacky since we normally use the alloy facade crate, but a reth
-# type expects a trait from this crate
+reth-revm = { workspace = true }
+reth-evm = { workspace = true }
+alloy = { workspace = true, features = ["consensus", "signer-local", "network"] }
+revm = "*"
```
emostov, line 32: todo: pin this

bin/exec/src/evm/builder.rs
```diff
@@ -0,0 +1,160 @@
+//! `revm` execution handlers.
```
emostov, line 1:
```suggestion
//! Evm builder that includes custom logic for zero gas fee transactions.
```

bin/exec/src/evm/mod.rs
```diff
@@ -0,0 +1,824 @@
...
+            let tx = Transaction::Eip1559(inner_tx);
+            TransactionSigned::new_unhashed(tx, signature)
+        };
+
+        (transaction_signed, signer.address(), gas_limit)
+    }
+
+    #[test]
+    fn evm_transact_with_account_creation() {
+        // let (header, db, provider) = setup();
```
emostov, line 228:
```suggestion
```

bin/exec/src/evm/mod.rs
```diff
@@ -0,0 +1,823 @@
...
+        let bundle_account = output.bundle.state.get(&signer_address).unwrap().clone();
+
+        // New account info is expected
+        assert_eq!(
+            bundle_account.info.unwrap(),
+            AccountInfo {
+                // Balance does not chance
+                balance: U256::from(user_balance),
+                nonce: user_account.nonce + 1,
+                code_hash: B256::from(hex!(
```
emostov, line 639: TODO use `.. Default::default()`

bin/exec/src/evm/builder.rs
```diff
@@ -0,0 +1,160 @@
...
+        let state_nonce = ctx
+            .evm
+            .inner
+            .journaled_state
+            .load_account(caller, &mut ctx.evm.inner.db)?
+            .data
+            .info
+            .nonce;
+
+        // Check that the transaction's nonce is correct
```
MaanavKhaitan, line 51: just wanna confirm, there are no other checks that we need to do here right?

bin/exec/src/evm/mod.rs
```diff
@@ -0,0 +1,823 @@
...
+            Account, Block, BlockBody, BlockWithSenders, EthereumHardfork, ForkCondition,
+            Transaction,
+        },
+    };
+    use reth_evm::execute::{
+        BatchExecutor, BlockExecutionError, BlockExecutorProvider, BlockValidationError,
+    };
+    use reth_provider::AccountReader;
+    use reth_revm::{database::StateProviderDatabase, test_utils::StateProviderTest};
+    // use revm::precompile::primitives::{AccountInfo, Bytecode, JumpTable, LegacyAnalyzedBytecode};
```
MaanavKhaitan, line 135:
```suggestion
```

bin/exec/src/evm/mod.rs
```diff
@@ -0,0 +1,823 @@
...
+        let account = result.state.get(&signer_address).unwrap();
+
+        assert_eq!(
+            account.info,
+            AccountInfo { balance: U256::ZERO, nonce: 1, ..Default::default() }
+        );
+
+        let mut status = AccountStatus::LoadedAsNotExisting;
+        status.insert(AccountStatus::Touched);
+        assert_eq!(account.status, status);
```
MaanavKhaitan, line 253: do we wanna add a quick check to verify that the status is not `Touched` before sending the tx

bin/exec/src/evm/mod.rs
```diff
@@ -0,0 +1,823 @@
...
+
+        evm_config.fill_tx_env(
+            evm.tx_mut(),
+            &transaction_signed,
+            transaction_signed.recover_signer().unwrap(),
+        );
+
+        let result = evm.transact().unwrap();
+
+        assert_eq!(result.result.gas_used(), 21080);
```
MaanavKhaitan, line 242: let's move 21080 to a var at the top of this test

bin/exec/src/evm/builder.rs
MaanavKhaitan: not sure what checks are done in the existing default handler in reth

MaanavKhaitan (commented):
Overall LGTM, a few comments. Will wait for Lucas to review too

bin/exec/src/evm/mod.rs
```diff
@@ -0,0 +1,823 @@
...
+            }
+        );
+        // Original account info is as expected
+        assert_eq!(
+            bundle_account2.original_info.unwrap(),
+            AccountInfo { balance: U256::from(user2_balance), nonce: 5, ..Default::default() }
+        );
+        assert_eq!(bundle_account2.status, DbAccountStatus::Changed);
+    }
+}
```
MaanavKhaitan, line 823: I think you mentioned this in slack: let's add a test for eip4844 tx? just to be safe

lucasege (commented):
So do I understand it correctly that txs still have gas and gas limits, but we just arent checking/deducting/crediting for any gas on user balances?

emostov:
> So do I understand it correctly that txs still have gas and gas limits, but we just arent checking/deducting/crediting for any gas on user balances?


@lucasege yes - I should add a test to show that txns with a low gas limit get denied and blocks with too many transactions gas wise fail

bin/exec/src/evm/builder.rs
emostov: I added a link to the default handler at the top of this function

bin/exec/src/evm/mod.rs
emostov: Best we can do is just show the account does not exist. This touched status only apply's to the changeset type

emostov: a2f24d261d453e0ffe94708ec995381d510306fe

emostov:
> > So do I understand it correctly that txs still have gas and gas limits, but we just arent checking/deducting/crediting for any gas on user balances?
> 
> @lucasege yes - I should add a test to show that txns with a low gas limit get denied and blocks with too many transactions gas wise fail

@lucasege I added tests to show the block gas limit functionality still works: d47bc5a6e3ae82dc2a9d11a831e1e1dde6f9a642

bin/exec/src/evm/mod.rs
emostov: [93905ca](https://github.com/InfinityVM/InfinityVM/pull/409/commits/93905ca257e453b28fad0e79d18cb0afde306b48)

bin/exec/src/evm/builder.rs
MaanavKhaitan: sounds great thanks

MaanavKhaitan (approved):
image

bin/exec/src/evm/mod.rs
```diff
@@ -0,0 +1,945 @@
...
+
+impl IvmEvmConfig {
+    /// Create a new instance of [Self].
+    pub const fn new(chain_spec: Arc<ChainSpec>) -> Self {
+        Self { eth: EthEvmConfig::new(chain_spec) }
+    }
+}
+
+impl ConfigureEvmEnv for IvmEvmConfig {
```
lucasege, line 38: nit: is there an upstream/fork reference we can reference for this similar to the `builder` file? e.g. some reference for what we're doing here

bin/exec/src/evm/mod.rs
```diff
@@ -0,0 +1,945 @@
...
+        let signer = LocalSigner::random();
+
+        transaction_with_signer(signer, 0)
+    }
+
+    fn transaction_with_signer(
+        signer: LocalSigner<SigningKey>,
+        nonce: u64,
+    ) -> (TransactionSigned, Address, u64) {
+        let gas_limit = EXACT_GAS_USED + 1_000;
```
lucasege, line 202: nit: any reason for this +1000?

bin/exec/src/evm/mod.rs
```diff
@@ -0,0 +1,945 @@
...
+            &transaction_signed,
+            transaction_signed.recover_signer().unwrap(),
+        );
+
+        let result = evm.transact().unwrap();
+        assert_eq!(result.result.gas_used(), EXACT_GAS_USED);
+        let account = result.state.get(&signer_address).unwrap();
+        assert_eq!(
+            account.info,
+            AccountInfo { balance: U256::ZERO, nonce: 1, ..Default::default() }
```
lucasege, line 281: clarifying question: why does nonce start at 1?

lucasege: I would have thought that this only increments after the tx is committed/processed, or is it processed somewhere earlier here?

bin/exec/src/evm/mod.rs
```diff
@@ -0,0 +1,945 @@
...
+
+        let exact_gas_used = EXACT_GAS_USED;
+
+        // We know this is the exact gas used
+        header.gas_used = exact_gas_used;
+        // And the expected receipts root
+        header.receipts_root =
+            B256::from(hex!("5240c13baa9d1e0d29a6c984ba919cb949d4c1a9ceb74060760c90e4d1fcd765"));
+
+        let user_account = Account { nonce: 420, balance: U256::ZERO, bytecode_hash: None };
```
lucasege, line 780: nice

lucasege (approved):
LGTM, few questions

bin/exec/src/evm/mod.rs
emostov: Wanted to differiantiate between gas limit and the gas used. Gas limit is used when initially checking tx and while gas used is what actually counts to the finaly block size

bin/exec/src/evm/mod.rs
emostov: Yes I can add that. For this one though I call thru to the `EthEvmConfig` for every method - it feels a little hacky but its because `IvmEvmConfig` we need a set of traits implemented on `IvmEvmConfig` and it so it happens that for this trait we want to fall back to Eth behavour

emostov: For revm I linked it specifically because you can't click through from the public docs, but this trait you should

bin/exec/src/evm/mod.rs
emostov: In ` let result = evm.transact()` we are effectively running the evm and returning a state diff within `result`. Here the account is actually the new state, we get it from `result.state.get(&signer_address).unwrap();`

emostov: So this nonce reflects the new state

PR #424, opened 2024-12-24
feat: Add proxy contracts

MaanavKhaitan:
# What

- Adds proxy contract for all the consumers: mock consumer, matching game consumer, clob consumer
- Adds `initialize()` function instead of constructors in consumer contracts, and adds proxy in deploy scripts

# Testing

`forge test`: Passes
`make test-all`: Passes


lucasege (approved):
Might need a clippy-allow too many args, but worth looking into the complaint to make sure the allow sol calls are correct.

This looks good, but worth noting that OZ recommends the UUPS proxy going forward vs. transparentupgradeable proxy https://www.rareskills.io/post/uups-proxy

I dont think it really matters for our use-case since its mostly for gas, but worth looking into.

MaanavKhaitan:
Yep I agree we can use UUPS in the future, just doesn't seem high prio rn because we don't care as much about gas

Issue #410, opened 2024-12-21
coproc: make relay queue persistent between restarts


emostov:
To improve the robustness of ordered job relaying, we need to make the relay queue persistent across shutdowns. The queue should be able to pick up where it left off if the machine the CN is running is rebooted. 

Additionally, after the first relay attempt, we should check the chain to ensure the job result has not yet been submitted to handle the edge case of a job result transaction making its way onchain while the CN is down.

PR #425, opened 2024-12-24
feat: configurable initial relay retries; async db read helpers

### LINKED ISSUE #411: coproc: make relay retries configurable
We need to make this value configurable via CLI 

https://github.com/InfinityVM/InfinityVM/blob/d42e4bbe2a809b985a6d5d5d29cab2ab1f7a2a36/crates/coproc/src/relayer.rs#L40

Once it is configurable, we will want to increase this value to a very high number in the testnet.

emostov:
Changes
- Rename `max_retries` to `dlq_max_retries`. 
- Introduce `initial_relay_max_retries`. We set a very high high default of 1,000 because for the initial exchange use case we cannot proceed with relaying the next job until the current job makes its way on chain. Closes #411 
- Add async methods for db reads that use spawn blocking to perform the reads. This isn't related to the new configuration param, but @Avi-D-coder noticed that we where doing blocking DB reads within tokio async code which is not ideal so figure I would quickly add the change here. The naming convention I used here is that I appended `_sync` to the existing functions, and left the new async version names the original names


crates/coproc/src/relayer.rs
```diff
@@ -37,7 +37,8 @@ use tracing::{error, info};
 
 /// Delay between retrying failed jobs, in milliseconds.
 const JOB_RETRY_DELAY_MS: u64 = 500;
-const JOB_RETRY_COUNT: usize = 3;
+/// Max duration between retries in `relay_job_result`.
+const JOB_RETRY_MAX_DELAY_MS: u64 = 30 * 1_000;
```
emostov, line 41: Semi arbitrary. Open to input here

emostov:
The failing test looks to be a persistent issue. Investigating

crates/coproc/src/cli.rs
```diff
@@ -124,9 +124,15 @@ struct Opts {
...
+    /// Max number of retries for relaying a job in the dead letter queue (DLQ) before giving up.
+    /// Note that each job in the DLQ is tried once before trying again.
     #[arg(long, default_value_t = 3)]
-    max_retries: usize,
+    dlq_max_retries: usize,
+
+    /// Max number of retries when initially attempting to relay a job. If a job fails this many
+    /// times it will be added to the DLQ.
+    #[arg(long, default_value_t = 1_000)]
+    initial_relay_max_retries: usize,
```
Avi-D-coder, line 135: Tiny NIT: if we are casting to u32 why not just make the config values u32

Avi-D-coder (approved):
LGTM once the CI passes.

lucasege:
Is there an issue for this? perhaps https://github.com/InfinityVM/InfinityVM/issues/410

Issue #363, opened 2024-11-19
update coprocessor docs and diagram


emostov:
Update the coprocessor docs section in the book and diagram to account for the changes from #355

emostov:
This still needs to be done, but likely can wait another month since we will add a few more things like a background process to queue incomplete jobs for execution or relaying

PR #426, opened 2024-12-24
feat: use strictly incrementing nonce to determine job relay order

### LINKED ISSUE #410: coproc: make relay queue persistent between restarts

emostov:
Closes #410, but not by using persistence. Instead we rely on the nonce to determine global ordering in the entire system

Since we no longer rely on the order a request is sent, we can refactor the job pipeline a bit. We introduce the notion of actors. The primary actor is the execution actor. We also refactor the relaying to a simple relay actor.

The execution actor receives jobs, executes them, and whenever the next to be relayed completes execution it sends them over a channel to an associated relay actor. 

The relay actor primarily exists so we don't have to block the execution actor on relaying. The relay actors channel acts as a queue for relaying and the relay actor guarantees that will not pop the next job off the channel until the current job has finished relaying. 

Note, the relay actor might be a little overkill because we could instead just spawn a new task every time we want to relay without blocking - since eth transaction themselves have a nonce, they should all still have a global ordering. However, to be cautious to start I figure it was better to wait until a transaction lands on chain before the relaying the next one. The main case I think its might be helpful is when we need to retry submitting a job: if we just spawned new tasks we wouldn't be able to ensure that the eth nonce in retry transactions is correct

Note, at the moment we don't have any way of removing actors. This should be fine for the initial use case, but should be improved in the future

Other changes:
- Create new tokio runtimes for clob and matching game server in e2e setup logic
- Move clob db operations into spawn blocking blocks
- Use tokio mpsc channel for db writer service

Follow up
- https://github.com/InfinityVM/gtx/issues/29
- https://github.com/InfinityVM/InfinityVM/issues/363


crates/coproc/src/job_executor.rs
```diff
@@ -94,52 +95,42 @@ where
     /// Create a new instance of [Self].
     pub const fn new(
         db: Arc<D>,
-        exec_queue_receiver: Receiver<Job>,
+        executor_rx: Receiver<ExecutorMsg>,
```
emostov, line 98: Renamed for consistency with the other channel sender/receivers

crates/coproc/src/relayer.rs
```diff
@@ -370,6 +221,133 @@ where
...
+    ) -> Self {
+        Self { writer_tx, relay_rx, job_relayer, initial_relay_max_retries }
+    }
+
+    /// Start the relay actor
+    async fn start(self) {
+        loop {
+            // TODO: for some reason recv_async was not working and never pulling from
+            // the channel. This is a hack, but I assume there is some other issue I am
+            // missing
```
emostov, line 300: Couldn't figure this one out. Using `try_recv` seems ok for now but lmk if anyone has suggestions

crates/coproc/src/relayer.rs
```diff
@@ -370,6 +221,133 @@ where
...
+                Err(_error) => {
+                    tokio::time::sleep(tokio::time::Duration::from_millis(25)).await;
+                    continue;
+                }
+                Ok(relay_msg) => relay_msg,
+            };
+
+            let job = match msg {
+                RelayMsg::Relay(job) => job,
+                // RelayMsg::Exit => return Ok(()),
```
emostov, line 315:
```suggestion
```

examples/clob/node/src/batcher.rs
```diff
@@ -81,29 +82,34 @@ where
...
-            db.view(|tx| tx.get::<ClobStateTable>(prev_state_index))??.ok_or_eyre("start_state")?.0;
-        tokio::task::yield_now().await;
-
-        let mut requests = Vec::with_capacity((end_index - start_index + 1) as usize);
-        for i in start_index..=end_index {
-            let r = db
-                .view(|tx| tx.get::<RequestTable>(i))??
-                .ok_or_eyre(format!("batcher: request {i}"))?
+        let db2 = db.clone();
+        let (state_borsh, requests_borsh) = tokio::task::spawn_blocking(move || {
```
emostov, line 93: I was debugging some issues that I thought might be caused by the clob blocking. I left these changes since we should be doing sync DB IO in spawn blocking anyways

examples/clob/node/src/engine.rs
```diff
@@ -57,11 +57,18 @@ where
             receiver.recv().await.ok_or_eyre("engine channel sender unexpected dropped")?;
 
         let request2 = request.clone();
-        db.update(|tx| {
-            tx.put::<GlobalIndexTable>(SEEN_GLOBAL_INDEX_KEY, global_index)?;
-            tx.put::<RequestTable>(global_index, RequestModel(request2))
+        let db2 = db.clone();
+        tokio::task::spawn_blocking(move || {
```
emostov, line 61: Same as with the matching game, this sync DB IO should be in a blocking thread

crates/coproc/src/relayer.rs
emostov: I haven't tried using a channel lib either, but thought it would be nice to stay consistent with flume everywhere altho not critical

Avi-D-coder: Why do we need a mpmc (flume) queue instead of a mpsc (tokio)?

Avi-D-coder: Using flume in async code can be a bit tricky.
cancel safety: https://github.com/zesterer/flume/issues/104 (note a linked bug https://github.com/n0-computer/iroh/pull/2536)
Perf https://github.com/zesterer/flume/issues/152

crates/coproc/src/execute.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// We assume that the caller spawns exactly one job actor per consumer.
+    async fn start(self, nonce: JobNonce) {
+        let mut next_job_to_submit = nonce;
+        let mut join_set = JoinSet::new();
+        let mut completed_tasks = ExecutedJobs::new();
+
+        loop {
+            tokio::select! {
+                // Handle a new job request by starting execution
+                new_job = self.rx.recv_async() => {
```
Avi-D-coder, line 80: Flume does not seem to be cancel safe and cannot be used with select (not sure about using it as async in general). See my other comment and this issue someone had:
https://github.com/n0-computer/iroh/issues/2415#issuecomment-2289069405

emostov: Oh great to know

crates/coproc/src/relayer.rs
emostov: I remove all the flume usage in this PR here: https://github.com/InfinityVM/InfinityVM/pull/426/commits/b9bf54076cd3ec00f7eca018ba67a11e68f96b37. However, still having issues with using `recv` (you can see the tests fail in the linked commit). When I go back to `try_recv`, things work again: https://github.com/InfinityVM/InfinityVM/pull/426/commits/418692c1e5f621a595ba1ee3f067855661e381da

emostov:
I update the DB writer service to use mpsc channel instead of flume: [5e86aab](https://github.com/InfinityVM/InfinityVM/pull/426/commits/5e86aab5d315d48bc1830c8eaffc175852a43692)

crates/coproc/src/execute.rs
```diff
@@ -0,0 +1,153 @@
...
+                                let (tx, executor_complete_rx) = oneshot::channel();
+                                executor_tx2.send_async((job, tx)).await.expect("executor pool send failed");
+
+                                // Return the executed job
+                                executor_complete_rx.await
+                            });
+                        },
+                        None => {
+                            warn!("execute actor channel closed, exiting");
+                            break
```
Avi-D-coder, line 102: This None means we have no more messages coming into the actor, we likely still have jobs in progress.
We don't want to shutdown the actor here because it will lose the jobs in progress.
We only should shutdown when all jobs are relayed and this queue is droped and empty.

Avi-D-coder (changes requested):

crates/coproc/src/execute.rs
```diff
@@ -0,0 +1,153 @@
...
+                                relay_tx.send(RelayMsg::Relay(job)).await.expect("relay actor send failed.");
+
+                                next_job_to_submit += 1;
+                                while let Some(next_job) = completed_tasks.remove(&next_job_to_submit) {
+                                    // Relay any directly subsequent jobs that have been completed
+                                    relay_tx.send(RelayMsg::Relay(next_job)).await.expect("relay actor send failed.");
+                                    next_job_to_submit += 1;
+                                }
+                            } else {
+                                // This is not the next job to relay, so just store it for later
```
MaanavKhaitan, line 128: I'm curious, what's the reason we wanna have this `completed_tasks` tracking here in the executor? doesn't the queue in the relay basically guarantee this already?

crates/coproc/src/intake.rs
```diff
@@ -127,23 +139,36 @@ where
...
             .await
             .expect("db writer broken");
 
-        if job.relay_strategy == RelayStrategy::Ordered {
-            let consumer = job
-                .consumer_address
+        let consumer_address: [u8; 20] =
+            job.consumer_address.clone().try_into().expect("caller must validate address length.");
+
+        let execution_tx = if !self.active_actors.contains_key(&consumer_address) {
```
MaanavKhaitan, line 149: so if the CN goes down, the GTX server would need to re-submit jobs to the CN that weren't able to be relayed?

Makefile
```diff
@@ -54,6 +54,7 @@ contracts:
 	cp contracts/out/MockConsumer.sol/MockConsumer.json crates/contracts/json
 	cp contracts/out/TransparentUpgradeableProxy.sol/TransparentUpgradeableProxy.json crates/contracts/json
 	cp contracts/out/ProxyAdmin.sol/ProxyAdmin.json crates/contracts/json
+	cp contracts/out/Consumer.sol/Consumer.json crates/contracts/json
```
MaanavKhaitan, line 57: I was thinking about whether we need to modify the `JobManager` contract to enforce strictly incrementing nonce. Lmk if you disagree:

For stateful jobs, we don't need to enforce this because this is implicit through the state hash check

For stateless jobs, we don't need to enforce this because it doesn't matter what order the results are relayed in? Is there an example of a stateless job which would require results to be posted in order? I'm guessing that even if it is we don't need to enforce it at the protocol-level and can enforce it an individual Consumer-level

MaanavKhaitan (commented):
Added a few comments

crates/coproc/src/execute.rs
Avi-D-coder: The queue to the relayer does not order the tasks.
The executor actors does. The relayer actor could and maybe should be replaced with spawn tokio task, it's stateless.

MaanavKhaitan: ah I see, makes sense

crates/coproc/src/relayer.rs
```diff
@@ -364,12 +217,144 @@ where
...
+                        self.job_relayer.clone(),
+                        self.writer_tx.clone(),
+                        self.initial_relay_max_retries,
+                    )
+                    .await;
+                }
+                RelayStrategy::Unordered => {
+                    // Don't wait for the job to make its way into a block before relaying the next
+                    // job.
+                    let fut = Self::relay_job_result(
```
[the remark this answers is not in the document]
emostov: Future, will update it

crates/coproc/src/intake.rs
```diff
@@ -90,27 +98,31 @@ where
...
             unsafe_skip_program_id_check,
+            execution_actor_spawner,
+            active_actors: Arc::new(DashMap::new()),
+            http_eth_rpc,
         }
     }
 
     /// Submits job, saves it in DB, and pushes on the exec queue.
+    ///
+    /// Caution: this assumes the consumer address has already been validated to be exactly 20
```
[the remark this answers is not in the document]
emostov: The gRPC `submit_job` endpoint does this right now. The other user is the events listener and that is trusted code.

crates/coproc/src/relayer.rs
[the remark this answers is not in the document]
emostov: We have 3 seperate things in this file: the relay chain client (JobRelayer), the actor (RelayActor), and the dlq relayer (RelayRetry). So wanted to keep the name general

crates/coproc/src/intake.rs
emostov: Yes. We can implement some extra logic in the CN to do this, but ultimately GTX needs to be in charge

Makefile
emostov: Yup, I agree. The only reason to add it for GTX is just to give stronger guarantees that jobs will get rejected if they don't follow ordering, but its not a safety thing

crates/coproc/src/execute.rs
emostov: I was going based off this sentence in the docs: https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.recv. 

> This method returns None if the channel has been closed and there are no remaining messages in the channel’s buffer. This indicates that no further values can ever be received from this Receiver. The channel is closed when all senders have been dropped, or when [close](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.close) is called.

In the context of the coproc, the current logic will never drop the sender so it should never return `None`. That being said, I think your initial logic of only exiting once the join set is empty is more future proof for refactors

emostov: Fixed here: [619f5d2](https://github.com/InfinityVM/InfinityVM/pull/426/commits/619f5d2c96e42ad090be1810932f5704ea755168)

crates/coproc/src/relayer.rs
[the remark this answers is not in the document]
emostov: [619f5d2](https://github.com/InfinityVM/InfinityVM/pull/426/commits/619f5d2c96e42ad090be1810932f5704ea755168)

crates/coproc/src/execute.rs
emostov: The main case I think the relay actor might be helpful is when we need to retry submitting a job: if we just spawned new tasks we wouldn't be able to ensure that the eth nonce in the retried transactions is correct

crates/coproc/src/relayer.rs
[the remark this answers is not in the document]
MaanavKhaitan: ah cool makes sense

crates/coproc/src/intake.rs
MaanavKhaitan: sounds good

Makefile
MaanavKhaitan: sounds good, I think we can leave this for now then

crates/coproc/src/relayer.rs
lucasege: how would you like to handle this for this PR? is this worth others taking a look to try to resolve or should we just merge as is?

lucasege (commented):
left some comments

crates/coproc/src/execute.rs
lucasege: yeah i think it makes sense to keep as a service since it's fairly minimal right now.

crates/coproc/src/relayer.rs
emostov: Will try and find some time to debug `recv` this week, but will merge with `try_recv` if I can't

Avi-D-coder (approved):

emostov:
Made an issue for the `recv` bug in the relayer channel: https://github.com/InfinityVM/InfinityVM/issues. 

Planning on merging once CI passes

PR #432, opened 2024-12-25
WIP: meta: programdb

mayuridev:
closes #314 (#315, #317)

Ignoring contracts for now, block submissions w/ firewall

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,6 +119,7 @@ significant_drop_in_scrutinee = "allow"
 significant_drop_tightening = "allow"
 
 [workspace.dependencies]
+ivm-remote-db = { path = "crates/remote-db" }
 # a
 anyhow = { version = "1", default-features = false }
 axum = { version = "0", features = ["http2", "tokio", "json"], default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ members = [
 	"crates/db",
 	"crates/eip4844",
 	"crates/mock-consumer",
+	"crates/remote-db",
 	"crates/scripts",
 	"crates/sdk/abi",
 	"crates/sdk/proto",
--- a/crates/coproc/Cargo.toml
+++ b/crates/coproc/Cargo.toml
@@ -26,6 +26,7 @@ prometheus = { workspace = true }
 axum = { workspace = true }
 flume = { workspace = true }
 num_cpus = { workspace = true }
+async-trait = "0.1.74"
 
 ivm-contracts = { workspace = true }
 ivm-abi = { workspace = true }
@@ -34,7 +35,13 @@ ivm-proto = { workspace = true }
 ivm-zkvm-executor = { workspace = true }
 ivm-tracing = { workspace = true }
 ivm-eip4844 = { workspace = true }
+ivm-remote-db = { workspace = true }
 
 [dev-dependencies]
 ivm-test-utils = { workspace = true }
 ivm-mock-consumer = { workspace = true }
+mock-consumer-programs = { workspace = true }
+ivm-db = { workspace = true }
+ivm-zkvm-executor = { workspace = true }
+async-trait = "0.1.74"
+mockall = "0.11.4"
--- a/crates/coproc/src/intake.rs
+++ b/crates/coproc/src/intake.rs
@@ -6,6 +6,7 @@
 
 use crate::{
     relayer::Relay,
+    remote_db::RemoteElfClientTrait,
     writer::{Write, WriterMsg},
 };
 use alloy::{
@@ -14,7 +15,10 @@ use alloy::{
     signers::{Signer, SignerSync},
 };
 use flume::Sender;
-use ivm_db::{get_elf, get_job, tables::Job};
+use ivm_db::{
+    get_elf, get_job,
+    tables::{ElfWithMeta, Job},
+};
 use ivm_proto::{JobStatus, JobStatusType, RelayStrategy, VmType};
 use ivm_zkvm_executor::service::ZkvmExecutorService;
 use reth_db::Database;
@@ -33,6 +37,9 @@ pub enum Error {
     /// ELF with given program ID already exists in DB
     #[error("elf with program ID {0} already exists")]
     ElfAlreadyExists(String),
+    /// Remote DB error
+    #[error("remote db error: {0}")]
+    RemoteDb(#[from] tonic::transport::Error),
     /// Job already exists in DB
     #[error("job already exists")]
     JobAlreadyExists,
@@ -65,6 +72,7 @@ pub struct IntakeHandlers<S, D> {
     writer_tx: Sender<WriterMsg>,
     relay_tx: Sender<Relay>,
     unsafe_skip_program_id_check: bool,
+    config: crate::config::Config,
 }
 
 impl<S, D> Clone for IntakeHandlers<S, D>
@@ -80,6 +88,7 @@ where
             writer_tx: self.writer_tx.clone(),
             relay_tx: self.relay_tx.clone(),
             unsafe_skip_program_id_check: self.unsafe_skip_program_id_check,
+            config: self.config.clone(),
         }
     }
 }
@@ -98,6 +107,7 @@ where
         writer_tx: Sender<WriterMsg>,
         relay_tx: Sender<Relay>,
         unsafe_skip_program_id_check: bool,
+        config: crate::config::Config,
     ) -> Self {
         Self {
             db,
@@ -107,6 +117,7 @@ where
             writer_tx,
             relay_tx,
             unsafe_skip_program_id_check,
+            config,
         }
     }
 
@@ -150,7 +161,7 @@ where
     }
 
     /// Submit program ELF, save it in DB, and return program ID.
-    pub fn submit_elf(
+    pub async fn submit_elf(
         &self,
         elf: Vec<u8>,
         vm_type: i32,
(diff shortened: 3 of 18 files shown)


crates/coproc/src/intake.rs
```diff
@@ -175,12 +182,33 @@ where
...
+            .send((
+                Write::Elf { vm_type, program_id: program_id.clone(), elf: elf.clone() },
+                Some(tx),
+            ))
             .expect("writer channel broken");
-        let _ = rx.blocking_recv();
+        let _ = rx.await;
+
+        // Store the ELF in remote DB
+        match crate::remote_db::RemoteElfClient::connect(&self.config.config.remote_db.endpoint)
```
emostov, line 197: This makes sense to me to still do ELF submission via the coprocessor node 👍

crates/coproc/src/job_executor.rs
```diff
@@ -130,17 +142,25 @@ where
...
         metrics: Arc<Metrics>,
         writer_tx: Sender<WriterMsg>,
         relay_tx: Sender<Relay>,
     ) -> Result<(), Error> {
+        let rt = tokio::runtime::Runtime::new().expect("Failed to create runtime");
         let writer_tx2 = writer_tx;
 
         while let Ok(mut job) = exec_queue_receiver.recv() {
-            let elf_with_meta = match Self::get_elf(&db, &mut job, &metrics, writer_tx2.clone()) {
+            let elf_with_meta = match rt.block_on(Self::get_elf(
```
emostov, line 157: Just as a heads up, I think we will likely want to move ELF fetching to the Intake service to keep external IO out of the job executor - we want the dedicated job executor threads to only have to deal with execution and not necessarily be blocked by external IO. Ok to leave this logic for now since I am doing decently big refactor of the whole job pipeline here https://github.com/InfinityVM/InfinityVM/pull/426

crates/coproc/src/job_executor.rs
```diff
@@ -162,27 +182,91 @@ where
...
+                        metrics.incr_job_err(&FailureReason::DbErrGetElf.to_string());
+                        job.status = JobStatus {
+                            status: JobStatusType::Failed as i32,
+                            failure_reason: Some(FailureReason::DbErrGetElf.to_string()),
+                            retries: 0,
+                        };
+                        writer_tx
+                            .send((Write::JobTable(job.clone()), None))
+                            .expect("db writer broken");
+                        Err(FailureReason::DbErrGetElf)
```
emostov, line 267:
```suggestion
                        return Err(FailureReason::DbErrGetElf);
```

crates/coproc/src/job_executor.rs
```diff
@@ -162,27 +182,91 @@ where
...
-                job.status = JobStatus {
-                    status: JobStatusType::Failed as i32,
-                    failure_reason: Some(FailureReason::MissingElf.to_string()),
-                    retries: 0,
-                };
-                writer_tx.send((Write::JobTable(job.clone()), None)).expect("db writer broken");
+                // Try remote DB
+                match crate::remote_db::RemoteElfClient::connect(&config.remote_db.endpoint).await {
+                    Ok(mut client) => {
+                        match client.get_elf(job.program_id.clone()).await {
```
emostov, line 201: I would just do a soft return of the client here and then do this match statement below. It just avoids some nesting and I think less nesting makes it easier to read

emostov: Otherwise this logic all looks good!

Cargo.toml
```diff
@@ -118,6 +119,7 @@ significant_drop_in_scrutinee = "allow"
 significant_drop_tightening = "allow"
 
 [workspace.dependencies]
+ivm-remote-db = { path = "crates/remote-db" }
```
emostov, line 122: just as a tiny nit, we have been putting all the local deps at the bottom of this file

crates/coproc/src/intake.rs
```diff
@@ -61,10 +83,7 @@ pub struct IntakeHandlers<S, D> {
     db: Arc<D>,
     exec_queue_sender: Sender<Job>,
     zk_executor: ZkvmExecutorService<S>,
-    max_da_per_job: usize,
-    writer_tx: Sender<WriterMsg>,
-    relay_tx: Sender<Relay>,
-    unsafe_skip_program_id_check: bool,
+    config: IntakeConfig,
```
emostov, line 86: What was the motivation for putting some of these values in `IntakeConfig` while leaving others just are args?

crates/coproc/src/job_executor.rs
emostov: If you do a soft return early with the client, you will need to do an explicit return here to exit early

crates/coproc/src/remote_db.rs
```diff
@@ -0,0 +1,81 @@
...
+        debug!("Connecting to remote ELF store at {}", endpoint);
+        let client = ElfStoreClient::connect(endpoint.to_string()).await?;
+        Ok(Self { client })
+    }
+}
+
+/// Trait defining the interface for remote ELF storage operations.
+/// This trait is used to abstract the remote storage implementation and enable mocking in tests.
+#[async_trait]
+pub trait RemoteElfClientTrait {
```
emostov, line 29: nice

proto/remote_db/v1/elf_store.proto
```diff
@@ -0,0 +1,29 @@
+syntax = "proto3";
+package remote_db.v1;
+
+service ElfStore {
```
emostov, line 4: We should get some input from other people, but I think the crate name should match the service name. So we could rename the crate to `elf_store`. `remote_db` is just kinda vague imho

crates/remote-db/build.rs
```diff
@@ -0,0 +1,4 @@
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    tonic_build::compile_protos("../../proto/remote_db/v1/elf_store.proto")?;
```
emostov, line 2: We where originally doing this for the CN protos, but we moved to using a manual script. This makes the build pipeline simpler since we don't need proto deps or building anywhere. The script is here https://github.com/InfinityVM/InfinityVM/blob/main/crates/scripts/src/proto_build.rs

What I would do is add a new `compile_protos` invocation to that script thats something like

```rust
    tonic_build::configure()
        .out_dir(out_dir)
        .compile_protos(
            &[
                "proto/remote_db/v1/elf_store.proto",
            ],
            &["proto"],
        )
        .unwrap();
```

(Not sure if that is exact code but should be something similiar

crates/remote-db/src/bin/server.rs
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    let pool = PgPoolOptions::new().max_connections(5).connect(&database_url).await?;
+
+    sqlx::migrate!("./migrations").run(&pool).await?;
+
+    let addr = "[::0]:50052".parse()?;
+    let service = ElfStoreService::new(pool);
+
+    println!("ElfStore server listening on {}", addr);
```
emostov, line 21: We can do in a follow up, but I would just add a simple tracing subscriber so we can use the tracing macros

crates/remote-db/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "ivm-remote-db"
+version = "0.1.0"
+edition = "2021"
+
+[package.metadata.cargo-machete]
+ignored = ["prost"]
+
+[[bin]]
+name = "server"
```
emostov, line 10: I would change to `ivm-elf-store` or whatever we land on for crate name

crates/remote-db/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+
+[[bin]]
+name = "test_client"
+path = "src/bin/test_client.rs"
+
+[dependencies]
+tonic = { workspace = true }
+tokio = { workspace = true }
+prost = { workspace = true }
+sqlx = { version = "0.7", features = ["runtime-tokio-rustls", "postgres", "migrate", "any", "macros"], default-features = false }
```
emostov, line 21: Jut for consistency, I would add this as a workspace dep

crates/coproc/src/intake.rs
emostov: Also just what was the general thought with adding these config structs?

crates/coproc/src/intake.rs
```diff
@@ -150,15 +157,15 @@ where
...
         &self,
         elf: Vec<u8>,
         vm_type: i32,
         program_id: Vec<u8>,
     ) -> Result<Vec<u8>, Error> {
         let vm_type = VmType::try_from(vm_type).map_err(|_| Error::InvalidVmType)?;
 
-        if !self.unsafe_skip_program_id_check {
+        if !self.config.unsafe_skip_program_id_check {
             let derived_program_id = self.zk_executor.create_elf(&elf, vm_type)?;
```
emostov, line 169: Since this is becoming `async` (and thus being executed by the tokio runtime), we will want to move the `elf.zk_executor.create_elf(&elf, vm_type)?` into `spawn_blocking` since it will take a bit. 

Just for context, we have to be really careful about tokio code since we don't want anything to block for more then ~100 microseconds. For example, I just merged a PR to make all the DB reads happen inside of spawn blocking since even those take too long.

emostov:
Something I forgot on initial review: the program id (verifying key) in the embedded DB should be trusted. But the whenever we receive an ELF submission or download an ELF we should check that the derived program id matches the claimed program id. This should be a very uncommon operation so it shouldn't be too inefficient

crates/coproc/src/intake.rs
mayuridev: config has some operation settinsg that might change (`max_da_per_job`, etc.) and shared resources across multiple operations (`writer_tx`, etc.) and the the constructor args are for core deps (db, zk executor) etc so thought it would be more modular --> NodeConfig at the top containing all settings and component specific configs for focused settings

crates/coproc/Cargo.toml
```diff
@@ -26,15 +26,24 @@ prometheus = { workspace = true }
...
+anyhow = { workspace = true }
 
 [dev-dependencies]
 ivm-test-utils = { workspace = true }
-ivm-mock-consumer = { workspace = true }
\ No newline at end of file
+ivm-mock-consumer = { workspace = true }
+mock-consumer-programs = { workspace = true }
+ivm-db = { workspace = true }
+ivm-zkvm-executor = { workspace = true }
```
emostov, line 47:
```suggestion
```
these are already in normal deps, which are included in dev deps. you only need to re-include in dev deps normally if there is some different feature gating you

crates/coproc/src/job_executor.rs
emostov: Since #426 is merged, I would like to move pulling the fetching of the elf to the execution actor so we don't slow down execution. The `exec_queue_receiver` will now need to hold the ELF in addition to the job

crates/elf-store/src/bin/server.rs
```diff
@@ -0,0 +1,29 @@
...
+
+    let database_url = env::var("DATABASE_URL").expect("DATABASE_URL must be set");
+
+    if !sqlx::Postgres::database_exists(&database_url).await? {
+        sqlx::Postgres::create_database(&database_url).await?;
+    }
+
+    let pool = PgPoolOptions::new().max_connections(5).connect(&database_url).await?;
+
+    sqlx::migrate!("./migrations").run(&pool).await?;
```
emostov, line 19: Does this get included in the binary? If so this is good but if its reading from disk at runtime we will have issues deploying since we will always need to colocate the file at this path

crates/coproc/src/intake.rs
emostov: Tbh not a huge fan unless its functional, but also not necessarily opposed. Curious what others think

crates/coproc/src/job_executor.rs
emostov: I think you need to merge in main to get the changes from #426

emostov:
This needs main merged in see thread https://github.com/InfinityVM/InfinityVM/pull/432#discussion_r1912580528

PR #435, opened 2024-12-30
chore(deps): update reth & alloy

emostov:
Update reth and alloy deps. There seems to be some implicit feature gating updates that will help ivm lib consumers build ivm deps and reth deps in the same workspace


Avi-D-coder (approved):
This was easy to review ;)

PR #442, opened 2025-01-08
chore: state root tests

emostov:


emostov:
This isn't the right approach - we are going to need tests that re-execute payload builder payloads in the engine tree

PR #443, opened 2025-01-08
chore(deps): use latest reth

emostov:
This is the first step in pulling latest engine in the node builder: https://github.com/paradigmxyz/reth/issues/13701

Changes:
- Bump alloy deps
- Bumps revm to https://github.com/bluealloy/revm/tree/release/v52

Also includes #444 and #445


bin/exec/src/evm/mod.rs
```diff
@@ -177,14 +175,7 @@ mod test {
         };
         let db = StateProviderTest::default();
 
-        let chain_spec = Arc::new(
```
emostov, line 180: consolidated chain spec creation to make it easier to tell we are using the same chain spec across tests.

bin/exec/src/evm/mod.rs
```diff
@@ -240,7 +231,7 @@ mod test {
 
         let result = evm.transact().unwrap();
 
-        assert_eq!(result.result.gas_used(), EXACT_GAS_USED);
+        assert_eq!(result.result.gas_used(), EXACT_GAS_USED + UNKNOWN_GAS_REGRESSION);
```
emostov, line 234: Tried to look through the `revm` diff, but couldn't find any reason this would change

bin/exec/src/evm/mod.rs
```diff
@@ -152,6 +144,8 @@ mod test {
 
     // Exact gas used by the transaction returned by `transaction_with_signer`.
     const EXACT_GAS_USED: u64 = 21080;
+    // TODO: WHY DO WE NEED THIS??
```
lucasege, line 147: add an issue?

lucasege (approved):
Looks good, lets just merge in main first

PR #445, opened 2025-01-08
refactor: use new reth engine

emostov:
Targets #444

This PR updates us to use the new engine. Not only is the new engine faster and easier to understand, it also unlocks dev/auto mine mode for us.

This now works!
```rust
cargo run -p ivm-exec -- node --dev --dev.block-time 1s
```


lucasege (approved):
ez

PR #451, opened 2025-01-13
feat(ivm-exec)!: loosen up txn validation to allow accounts with no balance

emostov:
We had an issue where transactions from accounts with not enough balance to cover gas where getting rejected by validation logic. The issue is that our validation logic was a strict superset of the reth eth validation logic. However, the reth eth validation logic includes logic to check the account balance against gas cost.

To fix this without forking, the best thing I could think of was essentially just copying the eth validation logic and making the modifications. The result here is the PR is quite big and the validation logic includes stuff for older forks. The tradeoff is that while we have more code that isn't neccesarily useful, we have less of a diff to the upstream version. I left in code comments that link to the upstream version and try to make the differences very obvious, including with inline comments

New types:
- `IvmValidatorInner` which maps to reth's `EthValidatorInner`
- `IvmTransactionValidatorBuilder` which maps to reth's `EthValidatorBuilder`


bin/exec/src/pool/validator.rs
```diff
@@ -43,49 +69,652 @@ impl IvmTransactionAllowConfig {
...
+        // Checks for nonce
+        if tx_nonce < account.nonce {
+            return TransactionValidationOutcome::Invalid(
+                transaction,
+                InvalidTransactionError::NonceNotConsistent { tx: tx_nonce, state: account.nonce }
+                    .into(),
+            )
+        }
+
+        // This is where balance is normally checked
```
lucasege, line 376: nit: do you think it's worth just keeping the code in here but commented out so it's clear without needing to ref the diff?

bin/exec/src/pool/validator.rs
```diff
@@ -43,49 +69,652 @@ impl IvmTransactionAllowConfig {
...
+            }
+        }
+
+        // Return the valid transaction
+        TransactionValidationOutcome::Valid {
+            // WARNING: this is meant to represent the current account balance. Since we don't
+            // require an account to have a balance, we hardcode a high account balance that we
+            // are confident will ensure this transaction passes downstream checks.
+            // Upstream logic: https://github.com/paradigmxyz/reth/blob/d761ac42f5e15c46c00db90ca1b5b6f7f62a4f7c/crates/transaction-pool/src/validate/eth.rs#L475
+            balance: U256::from(u64::MAX),
```
lucasege, line 430: this is the part i worry may cause issues or at least may just cause some weirdness, e.g. would this update any wallet's reflection of your balance to some crazy number, or would RPCs be thrown off? needs some testing

lucasege (approved):
1 nit and then just a comment

bin/exec/src/pool/validator.rs
emostov: I generally dislike commented out code since it feels messy and indicates some other tooling should be used. In this case I would argue the link is more accurate / better source of truth

Issue #450, opened 2025-01-10
ivm-exec: e2e test for payload building and engine block insertion


emostov:
We want e2e tests for creating blocks and adding them to the engine tree. This test should be able to catch state root diffs between the payload builder and the engine execution.

Ideally, we can also test the full tx submission stack, including the transaction submission RPC and the transaction pool logic.

Tests:

-  A block with multiple transactions from payload builder can be inserted into engine tree
- A transaction from a sender on the allow list with no account balance works
- A transaction from a sender not on the allow list is denied, regardless of balance
- A transaction with a `to` from the allow list is accepted, from a sender with no balance

PR #452, opened 2025-01-14
test(ivm-exec): blob & allow config e2e

emostov:
closes: https://github.com/InfinityVM/InfinityVM/issues/450

Changes
- Introduce `IvmNode` to implement `Node`, `NodeTypes`, and `NodeTypesWithEngine` against. This helps us configure a node for the node builder.
- Add e2e test for blobs
- Add e2e test for allow config

Follow ups
- [ ] https://github.com/InfinityVM/InfinityVM/issues/462


bin/exec/src/pool/validator.rs
emostov, line 1: This diff is mostly just from resolving merge conflicts from when https://github.com/InfinityVM/InfinityVM/pull/451/files squash merged into main

bin/exec/src/lib.rs
```diff
@@ -37,3 +55,81 @@ pub struct IvmCliExt {
...
+        EthereumNetworkBuilder,
+        IvmExecutorBuilder,
+        EthereumConsensusBuilder,
+    >;
+
+    type AddOns = IvmAddOns<
+        NodeAdapter<N, <Self::ComponentsBuilder as NodeComponentsBuilder<N>>::Components>,
+    >;
+
+    /// Denies all transactions by default
```
lucasege, line 127: What part of the code achieves this "denies by default"? I guess this comment is just to let you know under the hood that's the underlying behavior?

bin/exec/tests/e2e/allow.rs
```diff
@@ -0,0 +1,97 @@
...
+            .genesis(genesis.clone())
+            .cancun_activated()
+            .build(),
+    );
+    let node_config = NodeConfig::test()
+        .with_chain(chain_spec)
+        .with_unused_ports()
+        .with_rpc(RpcServerArgs::default().with_unused_ports().with_http());
+
+    // TODO: use non-funded addresses
```
lucasege, line 50: should we do this for this PR?
Maybe you can just create a new wallet from a random string?

bin/exec/tests/e2e/allow.rs
```diff
@@ -0,0 +1,97 @@
...
+    let node = NodeTestContext::new(node, eth_payload_attributes).await?;
+
+    let rpc = ProviderBuilder::new()
+        .with_recommended_fillers()
+        .wallet(EthereumWallet::new(alice_wallet.clone()))
+        .on_http(node.rpc_url());
+
+    let transfer_tx = TransactionTestContext::transfer_tx_bytes(1, alice_wallet.clone()).await;
+    let transfer_error = node.rpc.inject_tx(transfer_tx).await.unwrap_err();
+    assert_unsupported_tx(transfer_error);
```
lucasege, line 71: I don't really understand why the error here is an "unsupported tx"? I would expect that it is just a failed tx? is that just the specific way it was implemented if it gets dropped from the pool?

bin/exec/tests/e2e/blob.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    // inject normal tx
+    let raw_tx = TransactionTestContext::transfer_tx_bytes(1, second_wallet.clone()).await;
+    let tx_hash = node.rpc.inject_tx(raw_tx).await?;
+    // build payload with normal tx. This uses the node payload builder component
+    let (payload, attributes) = node.new_payload().await?;
+
+    // clean the pool
+    node.inner.pool().remove_transactions(vec![tx_hash]);
```
lucasege, line 84: why do we need to "clean the pool"? I would assume txs get cleared when a payload is created?

bin/exec/tests/e2e/blob.rs
```diff
@@ -0,0 +1,125 @@
...
+    // submit normal payload
+    node.engine_api.submit_payload(payload, attributes, PayloadStatusEnum::Valid).await?;
+
+    tokio::time::sleep(std::time::Duration::from_secs(3)).await;
+
+    // expects the blob tx to be back in the pool
+    let envelope = node.rpc.envelope_by_hash(blob_tx_hash).await?;
+    // make sure the sidecar is present
+    TransactionTestContext::validate_sidecar(envelope);
+
```
lucasege, line 123: Everything up to this chunk makes sense, this part I don't follow
Why are we sending two fcus back to back? Wouldn't it be possible to send one fcu that includes both?
Why is the blob tx back in the pool? 

The cause and effect here seems extremely random so probably relies on some internal understanding that is not explained yet. It's not clear to me how the blob payload is removed here and also why we would even want to do that

bin/exec/src/pool/validator.rs
lucasege: the diff is large, so if nothing here has changed from previous PRs then this has already been reviewed and is ok, otherwise lmk and I will comb through more carefully. a quick glance seemed familiar

lucasege (commented):
A few questions

bin/exec/tests/e2e/blob.rs
emostov: Creating a new payload does not clear the transaction pool. Clearing the transaction pool, from my understanding, is a function of the canonical chain (so blocks specified from fork choice updates)

bin/exec/src/lib.rs
emostov: This is a dated comment from when I needed a default impl. We now have the type hold a `IvmTransactionAllowConfig` so it needs to be initialized with an allow config and does not come with a default impl

bin/exec/tests/e2e/allow.rs
emostov: That is the specific error we return for non-allowed transactions. I chose this because it was simpler to re-use an existing error instead of figuring out how to add a custom error - I wanted to clearly document that in the tests. I can work on adding a custom error to try and make this special error case more clear

bin/exec/tests/e2e/allow.rs
emostov: My bad this is a dated comment when I still wasn't sure if the account has funding. The address is not funded - I believe this is because its not included in the genesis alloc

We show it has no funding below with `get_account_err = rpc.get_account(alice_wallet.address())`

bin/exec/tests/e2e/blob.rs
emostov: Check out the latest comments and slight change I have pushed. I think they will help explain

bin/exec/tests/e2e/allow.rs
lucasege: I think it works, my only concern is just that it will confuse users.
Maybe we can create an issue for longer term to create a custom error? probably not a mainnet blocker, but it would be nice eventually

bin/exec/tests/e2e/blob.rs
```diff
@@ -74,49 +86,49 @@ async fn can_handle_blobs() -> eyre::Result<()> {
...
     TransactionTestContext::validate_sidecar(envelope);
 
-    // build a payload
+    // Build blob payload
     let (blob_payload, blob_attr) = node.new_payload().await?;
 
-    // submit the blob payload
+    // Submit the blob payload. At this point we have only submitted the blob payload
     let blob_block_hash =
         node.engine_api.submit_payload(blob_payload, blob_attr, PayloadStatusEnum::Valid).await?;
```
lucasege, line 113: These comments make more sense now

Nit: what exactly does `submit_payload` do? I can understand building the payload and then FCUing it, but i'm not 100% sure what submitting means

lucasege (approved):
These comments make sense now. Left one nit but all looks good

PR #454, opened 2025-01-15
test: Add e2e test for parallel job submission

0x25d310b7:
This test demonstrates an issue with job submission under concurrency. Before we roll out a fix, I wanted to prepare a PR that shows the problem. 

In the majority of test runs, the coprocessor stalls after relaying the first transaction. The reason is that we spawn one execution actor per gRPC call, when it should be only a single one.

### DIFF
--- a/test/e2e/src/lib.rs
+++ b/test/e2e/src/lib.rs
@@ -36,6 +36,8 @@ pub struct Args {
     pub anvil: AnvilJobManager,
     /// Coprocessor Node gRPC client.
     pub coprocessor_node: CoprocessorNodeClient<Channel>,
+    /// Coprocessor Node gRPC endpoint.
+    pub coprocessor_node_endpoint: String,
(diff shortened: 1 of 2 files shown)


emostov:
Confirming tests fail with this warn log repeating

Screenshot 2025-01-14 at 7 16 35 PM

emostov (approved):
Test looks great 🥇 . I confirmed the log line indicates the sending side of the executor actor is getting dropped and in this context that means there are many actors where there should only be one and we are losing the send side for each actor

https://github.com/InfinityVM/InfinityVM/blob/zeke-blob-e2e-test/crates/coproc/src/execute.rs#L101

0x25d310b7:
Thanks @emostov! I'm going to close this PR in favor of https://github.com/InfinityVM/InfinityVM/pull/455, which combines the fix and test.

PR #455, opened 2025-01-15
fix: Fix race condition in active_actors map

0x25d310b7:
The surface-level problem is that the `contains` check and the `insert` in the intake handler are not atomic. Because we hold no lock on the hashmap, multiple grpc calls are able to pass the `contains` check before any of them reach the `insert`. In the end we end up with multiple actors per consumer address, which breaks downstream code.

These kinds of problems are generally solved with guard objects that you get via e.g. entry-style APIs. Unfortunately, if you hold dashmap's mapref across an await point, you can run into a deadlock when another task calls the blocking `entry` function.

This PR fixes this by adding an entry call, and lifting the await code. I merged in the test case from https://github.com/InfinityVM/InfinityVM/pull/454, which should now be passing.

### DIFF
--- a/crates/coproc/src/intake.rs
+++ b/crates/coproc/src/intake.rs
@@ -16,14 +16,12 @@ use alloy::{
     signers::{Signer, SignerSync},
     transports::http::reqwest::Url,
 };
+use dashmap::{DashMap, Entry};
 use ivm_db::{get_elf_sync, get_job, tables::Job};
 use ivm_proto::{JobStatus, JobStatusType, VmType};
 use ivm_zkvm_executor::service::ZkvmExecutorService;
 use reth_db::Database;
-use std::{
-    collections::{hash_map::Entry, HashMap},
-    sync::Arc,
-};
+use std::sync::Arc;
 use tokio::sync::{mpsc::Sender, oneshot};
 
 /// Errors from job processor
@@ -72,7 +70,7 @@ pub struct IntakeHandlers<S, D> {
     writer_tx: Sender<WriterMsg>,
     unsafe_skip_program_id_check: bool,
     execution_actor_spawner: ExecutionActorSpawner,
-    active_actors: Arc<tokio::sync::Mutex<HashMap<[u8; 20], Sender<Job>>>>,
+    active_actors: Arc<DashMap<[u8; 20], Sender<Job>>>,
     http_eth_rpc: Url,
 }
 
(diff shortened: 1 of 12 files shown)


emostov (approved):
This makes sense to me. Just want to double check the tokio async mutex is a good approach with @Avi-D-coder

Avi-D-coder (approved):
Your explanation here is great.
The change is good, my one concern is the use of a global `tokio::mutex`.

I'd prefer `tokio:RwLock`, with a read precheck to minimize contention.

Avi-D-coder (changes requested):
Wait can't we lift the call to `getNextNonce` in to a precheck and use a dashmap entry to make it atomic.

Basically:

```
loop {
  if !self.active_actors.contains_key {
    let nonce = consumer.getNextNonce().call().await?._0;
    // dashmap entry lock
    match self.active_actors.entry {
      Occupied => continue,
      Entry::Vacant => current logic
   }
  }
  
  ```
  
This logic is still wrong because there's a race between sending the job to the actor and actor deletion so you need to use a entry lock in the contains path too. Note that the race actually exists in the mutex code too. 

The advantage of this would be more granular locking, but the rwlock version where you lock while sending the job would be fine too.

Although I don't think we ever currently delete actors, we will need to eventually.

**Comment by 0x25d310b7** [comment]
Thanks for the suggestion. I added dashmap back, and lifted the async above the entry call. I don't think a loop should be necessary here though. PTAL

File: crates/coproc/src/intake.rs [hunk 1]
```diff
@@ -147,23 +147,26 @@ where
...
-        let execution_tx = if !self.active_actors.contains_key(&consumer_address) {
-            // Get the latest nonce to initialize the execution actor with
-            let provider = ProviderBuilder::new().on_http(self.http_eth_rpc.clone());
-            let consumer =
-                ivm_contracts::consumer::Consumer::new(consumer_address.into(), provider);
-            // This is the next nonce; since we expect contracts to be initialized with nonce 0, the
-            // first nonce should be 1.
-            let nonce = consumer.getNextNonce().call().await?._0;
+        let provider = ProviderBuilder::new().on_http(self.http_eth_rpc.clone());
+        let consumer = ivm_contracts::consumer::Consumer::new(consumer_address.into(), provider);
```
Avi-D-coder, line 151: NIT: move these out of the happy path down into the else clause.

Avi-D-coder (approved):

Avi-D-coder:
I have one NIT, but other than that this is good. 

If we ever make it so actors can be removed we will have to hold the lock while we send. To avoid the potential race. Which would mean we would also have to use `try_send` and then that gets into backpressure yada yada yada.

Avi-D-coder:
@emostov how do you feel about this approach?

crates/coproc/src/intake.rs
0x25d310b7: done!

Avi-D-coder (approved):

crates/coproc/src/intake.rs
```diff
@@ -147,26 +147,31 @@ where
...
-        } else {
-            self.active_actors
-                .get(&consumer_address)
-                .expect("we checked above that the entry exists. qed.")
-                .clone()
+            match self.active_actors.entry(consumer_address) {
+                Entry::Occupied(e) => e.get().clone(),
+                Entry::Vacant(e) => {
+                    // ATTENTION: No async allowed while an entry lock is held.
+                    let execution_tx = self.execution_actor_spawner.spawn(nonce);
```
emostov, line 165: Technically we are spawning a new tokio task - not sure if that counts as async

https://github.com/InfinityVM/InfinityVM/blob/477db215faa029af1737af7fa5cccb165d4258b4/crates/coproc/src/execute.rs#L54

Avi-D-coder: Good catch, no it's not.
That should be lifted above next to nonce.
If we then find an Occupied entry the new actor will close when it's tx   channel is dropped.

emostov: And in the relay actor the exit msg does indeed exit:

https://github.com/InfinityVM/InfinityVM/blob/477db215faa029af1737af7fa5cccb165d4258b4/crates/coproc/src/relayer.rs#L319-L322

Avi-D-coder: @0x25d310b7 did this get fixed?

0x25d310b7: Discussed on slack. Spawn doesn't yield to the runtime. As an additional confirmation, we have the docs of tokio::spawn essentially saying the same:

> It is guaranteed that spawn will not synchronously poll the task being spawned. This means that calling spawn while holding a lock does not pose a risk of deadlocking with the spawned task.

https://docs.rs/tokio/latest/tokio/task/fn.spawn.html

Issue #449, opened 2025-01-10
ivm-exec: Adjusting tx priority to be FIFO


emostov:
Since we do not charge gas for transactions, we need to ignore priority fees since those can be set to an arbitrary number regardless of a users balance. Ideally we should support FIFO. If we can't due that we can randomize it or assign equal priority to all transactions. 

We might want some way to prioritize coproc job results, but that is likely out of scope for an initial impl.

Default payload builder takes a best_txs: https://github.com/paradigmxyz/reth/blob/ca30702f85768d75722260c37907f3ceec35c577/crates/ethereum/payload/src/lib.rs#L172

We may also want some way to prioritize allow listed addresses

PR #466, opened 2025-01-18
feat: add gasless transaction priority 

emostov:
closes: https://github.com/InfinityVM/InfinityVM/issues/449
See in code docs


lucasege (approved):
This seems reasonable to me. the only question I have is if there are any ways for transactions to get stuck. It seems like the base fee question is one, we'll need to review how base fees are adjusted in the eip1559 logic and if there are any ways to make sure this can't happen

PR #472, opened 2025-01-21
feat(ivm-exec)!: fork aware allow list and priority senders

emostov:
BREAKING CHANGE: this is a breaking change for the ivm_config.toml. Existing ivm_config.tomls should migrate to the new format and change their current allowed rules to have timestamp 0.

This PR introduces two major changes:

- The ivm_config.toml can now specify priority senders. These will take strict precedence over all other senders when constructing payloads. These should be senders making system critical transactions. This is not consensus critical and can be different between nodes.
- The allow config is now determined by block timestamp. This allows creating new allow configs that will be enacted at some future time in the future. We may hardcode these values into mainnet binaries so operators do not need to worry. For now they need to be specified by a file. You can override this for develop with `--tx-allow.all`

Follow up:
-[ ] bake in configs for live networks


bin/exec/Cargo.toml
```diff
@@ -45,19 +45,25 @@ toml = { workspace = true }
...
 alloy-sol-types = { workspace = true }
 alloy-contract = { workspace = true }
+alloy-rpc-types-eth = { workspace = true }
+
+reth-provider = { workspace = true, features = ["test-utils"] }
+reth-e2e-test-utils = { workspace = true }
+reth-node-core = { workspace = true }
+
+# enable this crates own test utils feature
+ivm-exec = { path = ".", features = ["test-utils"] }
```
emostov, line 62: we enable our own test-utils to ensure e2e tests in this crate have access to them

bin/exec/src/config/transaction.rs
emostov, line 1: This logic used to be in validator.rs. I moved it here and added unit tests

bin/exec/src/config/mod.rs
```diff
@@ -0,0 +1,408 @@
...
+use alloy_primitives::{map::foldhash::HashSet, Address};
+use eyre::eyre;
+use std::{collections::BTreeMap, fs, path::Path};
+use transaction::IvmTransactionAllowConfig;
+
+pub mod transaction;
+
+/// IVM specific configuration for the execution client. This is the in memory representation.
+///
+/// N.B. The default will allow all transactions.
```
lucasege, line 12: nit: similar question on default being allow vs. restrict

lucasege (approved):
This looks good and testing looks comprehensive. a few small questions
Also, it would be helpful to run this by @twoshark before we deploy it so he is familiar with the forking pattern and can give input on any slight changes we may want

bin/exec/README.md
```diff
@@ -16,19 +17,62 @@ The node has a standard reth CLI and uses the reth toml config.
 
 #### Allow lists
 
-The allow lists can be configured in an IVM config toml. If no config is found at start up, a restrictive default will get created in the data dir at `<DATADIR>/ivm_config.toml`. To specify a config at a different path you can run `cargo run --bin ivm-exec -- node --ivm-config <PATH>`.
+The allow lists can be configured in an IVM config toml. If no config is found at start up, a permissive default will get created in the data dir at `<DATADIR>/ivm_config.toml`. To specify a config at a different path you can run `cargo run --bin ivm-exec -- node --ivm-config <PATH>`. In the future we will likely add defaults to the binary for major networks.
```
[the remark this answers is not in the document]
emostov: Ya I can change this back to being restrictive. In practice I expect us to always have a config, I just wanted to make it easier to start a dev node.

emostov:
> This looks good and testing looks comprehensive. a few small questions Also, it would be helpful to run this by @twoshark before we deploy it so he is familiar with the forking pattern and can give input on any slight changes we may want

Would be great to coordinate with @twoshark so I can bake monza and suzuka configs in

PR #476, opened 2025-01-22
feat(coproc): compress offchain input before blob enconding

### LINKED ISSUE #475: coproc: Compress offchain input before blob encoding
Probably want to look at https://github.com/pseitz/lz4_flex or https://github.com/BurntSushi/rust-snappy

We would want to encode right before this step: https://github.com/InfinityVM/InfinityVM/blob/0dd6d02b596f6934f85600ad25e18c89db7eb213/crates/zkvm-executor/src/service.rs#L141

emostov:
Closes #475 

This is a breaking change for any one consuming coproc blob data

I chose lz4 because, from what I understood, the compressed data will likely be smaller then snappy

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4991,6 +4991,7 @@ dependencies = [
  "ivm-eip4844",
  "ivm-proto",
  "ivm-zkvm",
+ "lz4_flex",
  "thiserror 1.0.69",
  "tracing",
 ]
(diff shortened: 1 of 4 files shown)


Avi-D-coder (approved):
Lgtm, I am surprised this doesn't affect any of the tests?

emostov:
> Lgtm, I am surprised this doesn't affect any of the tests?

We don't decode the blob data anywhere, so there are no consumers to update

PR #477, opened 2025-01-22
fix(coproc): validate compressed data length

emostov:
Follow up to #476. We want to validate compressed data length, else risk rejecting offchain input that would be valid once compressed.


0x25d310b7 (approved):
LGTM

PR #482, opened 2025-01-23
feat(coproc): pending jobs endpoint

emostov:
This adds a get pending jobs gRPC endpoint. The endpoint will return the jobs the coprocessor currently has pending.

Implementation:

The execution actor tracks all job nonces the moment they come in. When a pending job request is the made, the intake will first query the chain for the `next_nonce` and then pass that to the execution actor. The execution actor will then filter its pending jobs collection by removing any nonces lower then the `next_nonce`, since we know those have been relayed successfuly. That collection is then returned to the user through the call stack.

The pending job collection has the potential to to become a memory leak if the jobs never relay because the actor will never exit.

Other changes:
- move proto gen script to its own crate so we don't have to worry about proto generation breaking the script itself.
- update CONTRIBUTING.md
- add proto gen make command alias

TODO:
- don't track unordered jobs.


crates/coproc/src/execute.rs
```diff
@@ -57,16 +57,27 @@ impl ExecutionActorSpawner {
...
 }
 
+/// A message to the execution actor.
+#[derive(Debug)]
+#[allow(clippy::large_enum_variant)]
+pub enum ExecMsg {
+    /// Send a job to execute and relay.
+    Exec(Job),
+    /// Request the current pending jobs. The given job nonce is the highest
+    /// job onchain.
```
emostov, line 67:
```suggestion
    /// Request the current pending jobs. The given job nonce is expected
    /// to be the next nonce on chain.
```

test/e2e/tests/pending_jobs.rs
```diff
@@ -0,0 +1,232 @@
...
+        futures::future::join_all(futures).await;
+
+        let pending = args
+            .coprocessor_node
+            .get_pending_jobs(pending_jobs_request.clone())
+            .await
+            .unwrap()
+            .into_inner();
+        assert_eq!(pending.pending_jobs, vec![1, 2, 3]);
+        dbg!(pending.pending_jobs);
```
emostov, line 90:
```suggestion
```

emostov:
superseded by #483

PR #483, opened 2025-01-23
feat(coproc): pending jobs endpoint

emostov:
Closes https://github.com/InfinityVM/gtx/issues/219

This adds a get pending jobs gRPC endpoint. The endpoint will return the jobs the coprocessor currently has pending.

Implementation:

The execution actor tracks all job nonces the moment they come in. The relayer sends message to execution actor to indicate when a job has been relayed.

PR #484, opened 2025-01-24
chore: move proto gen script to its own crate

emostov:
Changes
- move proto gen script to its own crate so we don't have to worry about proto generation breaking the script itself.
- update CONTRIBUTING.md
- add proto gen make command alias

This is a backport from work #483


CONTRIBUTING.md
```diff
@@ -6,14 +6,6 @@ This repo is a rust workspace. The workspace configuration is defined by `Cargo.
 
 ### Install tools
 
-Rust
```
MaanavKhaitan, line 9: why remove the rust instructions?

MaanavKhaitan (approved):
LGTM, I don't have a lot of context on the proto stuff though so trusting you know this better

CONTRIBUTING.md
emostov: We now have https://github.com/InfinityVM/InfinityVM/blob/main/rust-toolchain.toml which should tell peoples rustup what version to use

Issue #479, opened 2025-01-22
coproc: Jobs need to be re-trigger-able so they can get unstuck


emostov:
If a job can get stuck in a state where it has not been relayed. We need a way to be able to pick those jobs back up and keep processing them

PR #491, opened 2025-01-25
feat: allow retriggering jobs

emostov:
closes: #479


crates/coproc/src/execute.rs
```diff
@@ -103,15 +105,26 @@ impl ExecutionActor {
                 new_job = rx.recv() => {
                     match new_job {
                         Some(ExecMsg::Exec(job)) => {
+                            if !job.is_pending() && !job.is_done() || (job.is_ordered() && job.nonce <  next_job_to_submit) {
```
emostov, line 108: In the case of `job.nonce <  next_job_to_submit` there is no way to return the error to the consumer at this point unless we come up with a new status type for a submitted job and then they could see it through the get result endpoint.

I was thinking of checking this condition in intake, but a query to the chain is kind of expensive. The other option is allowing this actor to be queried for the next nonce, but that seemed like to much complexity for this PR

crates/coproc/src/intake.rs
```diff
@@ -136,23 +142,43 @@ where
...
+        let consumer_address: [u8; 20] =
+            job.consumer_address.clone().try_into().expect("caller must validate address length.");
+
         // TODO: add new table for just job ID so we can avoid writing full job here and reading.
         // We can just pass the job itself along the channel
         // full job https://github.com/InfinityVM/InfinityVM/issues/354
-        if get_job(self.db.clone(), job.id).await?.is_some() {
-            return Err(Error::JobAlreadyExists);
-        }
+        let job = if let Some(old_job) = get_job(self.db.clone(), job.id).await? {
```
0x25d310b7, line 151: Not blocking, but I think there's a race between this get_job check and the write to the db, right? We might end up submitting multiple jobs with the same id for execution, but the job table would only have one entry.

0x25d310b7 (approved):
Nice! LGTM, but had one non-blocking comment.

crates/coproc/src/execute.rs
0x25d310b7: Big picture-wise, I would favor doing checks in intake and keeping the code in the executor simpler. E.g. right now we are basically doing an equivalent of the `!job.is_pending() && !job.is_done()` check in intake (we don't send if relayed/failed). 

So maybe querying the actor for the next nonce would be better. Then there'd be no need for this whole check.

emostov: I agree with your point. I think I will quickly refactor to query the actor for the nonce check

crates/coproc/src/intake.rs
emostov: Thats a good point. I'm not sure how to prevent that with the current design. I assume there is some one we could ID jobs by hash, but not sure exactly how that could work

Issue #423, opened 2024-12-23
coproc: ensure buffered DB writes complete on graceful shutdown


emostov:
The coproc has a thread dedicated to DB writes. Writes are buffered over a channel to the DB. To improve resillency, we the writer should wait until all buffered writes have completed before exiting.

At the moment, we just exit when we receive a kill message

https://github.com/InfinityVM/InfinityVM/blob/d42e4bbe2a809b985a6d5d5d29cab2ab1f7a2a36/crates/coproc/src/writer.rs#L95

PR #492, opened 2025-01-26
chore: batch db writes and flush writes before exit

emostov:
closes https://github.com/InfinityVM/InfinityVM/issues/423

### DIFF
--- a/crates/coproc/src/writer.rs
+++ b/crates/coproc/src/writer.rs
@@ -11,6 +11,9 @@ use reth_db::{
 use std::sync::Arc;
 use tokio::sync::{mpsc::Receiver, oneshot};
 
+const WRITE_BUFFER_SIZE: usize = 256;
+const LAST_WRITE_BUFFER_SIZE: usize = 10_240;
+
 /// A write request to the [`Writer`]. If a sender is included, the writer
 /// will respond once the write has been completed.
 pub type WriterMsg = (Write, Option<oneshot::Sender<()>>);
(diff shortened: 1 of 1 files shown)


crates/coproc/src/writer.rs
```diff
@@ -11,6 +11,9 @@ use reth_db::{
 use std::sync::Arc;
 use tokio::sync::{mpsc::Receiver, oneshot};
 
+const WRITE_BUFFER_SIZE: usize = 256;
+const LAST_WRITE_BUFFER_SIZE: usize = 10_240;
```
emostov, line 15: These numbers are arbitrary. I wanted to have the last write make a best effort to complete all the writes

crates/coproc/src/writer.rs
```diff
@@ -11,6 +11,10 @@ use reth_db::{
 use std::sync::Arc;
 use tokio::sync::{mpsc::Receiver, oneshot};
 
+// These numbers are arbitrary.
+const BATCH_SIZE: usize = 64;
+const KILL_BATCH_SIZE: usize = 4 * BATCH_SIZE;
```
Avi-D-coder, line 16: I don't understand what `KILL_BATCH_SIZE` means?

crates/coproc/src/writer.rs
```diff
@@ -74,9 +78,31 @@ where
...
+        while rx.blocking_recv_many(&mut buffer, BATCH_SIZE) > 0 {
+            let kill_seen = Self::write_batch(&db, &mut buffer)?;
+            if kill_seen {
+                break
+            }
+        }
+
+        if rx.blocking_recv_many(&mut buffer, KILL_BATCH_SIZE) > 0 {
+            Self::write_batch(&db, &mut buffer)?;
+        }
```
Avi-D-coder, line 93: I need to look at the code that sends the kill signal, I have not yet, but just going off these lines I don't understand why we know all messages have been written after an additional `KILL_BATCH_SIZE` have been read.

After I look at the sender I am sure it'll make more sense, but it'd be nice to have a could of comments here explaining.

emostov: We actually don't even have a kill signal handler yet. The full graceful shutdown flow needs more thought. Right now it just sends the kill message if any of the handles it try joins on errors. LMK if you have recommendations

https://github.com/InfinityVM/InfinityVM/blob/cce718cf5baaf3f3168e4f23645b4319a0225d70/crates/coproc/src/node.rs#L265-L279

PR #493, opened 2025-01-28
feat: update to SP1 4.0.1 & use executor directly

Avi-D-coder:
This PR is needed for some of the profiling work to not build two versions of SP1, but it seems to introduce a major perf regresion in the tests.

I will profile in follow up work or before this is merged if people would like.

### DIFF
--- a/examples/clob/programs/program/Cargo.toml
+++ b/examples/clob/programs/program/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2021"
 [workspace]
 
 [dependencies]
-sp1-zkvm = { git = "https://github.com/succinctlabs/sp1.git", rev = "15be73d3acae3718a0acc529ed566f6a448b1c0e", features = ["libm", "lib"], default-features = false }
+sp1-zkvm = { version = "4", features = ["libm", "lib", "embedded"], default-features = false }
 alloy = { version = "0.8", features = ["sol-types"], default-features = false }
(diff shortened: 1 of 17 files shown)


examples/clob/programs/program/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2021"
 [workspace]
 
 [dependencies]
-sp1-zkvm = { git = "https://github.com/succinctlabs/sp1.git", rev = "15be73d3acae3718a0acc529ed566f6a448b1c0e", features = ["libm", "lib"], default-features = false }
+sp1-zkvm = { version = "4", features = ["libm", "lib", "embedded"], default-features = false }
```
emostov, line 9: is embedded prefered over the bump allocator

emostov (approved):

examples/clob/programs/program/Cargo.toml
Avi-D-coder: I don't know I put bump because my hunch is that in our execute only use case lower memory usage will be better than never freeing, but we should profile it and also see just how much memory is used by bump with large memory heavy coproc jobs.

We probably need to decide on a limit on memory usage so a large job cannot cause an OOM.

emostov: Last time I checked they had a hardcoded limit of 64mb

Issue #338, opened 2024-11-05
coprocessor-node: persist `sp1_core_executor::Program` in db


emostov:
For perf, we should persist [sp1_core_executor::Program](https://docs.rs/sp1-core-executor/latest/sp1_core_executor/struct.Program.html) in the db so we don't have to construct it on the fly.

We should be able to use https://docs.rs/sp1-core-executor/latest/sp1_core_executor/struct.Program.html#method.from to create this when we receive an elf.

Lets also take out the verifying key check when executing. We are assuming we trust the embedded database

emostov:
We will want something like 

```
let program = Program::from(elf_bytes).expect("Failed to create program from ELF bytes");
```

PR #495, opened 2025-01-28
perf: persist `Program` program and use in executor

emostov:
Closes https://github.com/InfinityVM/InfinityVM/issues/338

There is still is room to do even better by keeping the program in memory and not double serializing it in the DB, but it keeps the executor trait api simpler if we just accept bytes and deserialize within the execute fn impl

### DIFF
--- a/crates/coproc/src/intake.rs
+++ b/crates/coproc/src/intake.rs
@@ -263,11 +263,25 @@ where
         }
 
         // Write the elf and make sure it completes before responding to the user.
-        let (tx, rx) = oneshot::channel();
         self.writer_tx
-            .blocking_send((Write::Elf { vm_type, program_id: program_id.clone(), elf }, Some(tx)))
+            .blocking_send((
+                Write::Elf { vm_type, program_id: program_id.clone(), elf: elf.clone() },
+                None,
+            ))
             .expect("writer channel broken");
-        let _ = rx.blocking_recv();
+
+        let program_bytes = self.zk_executor.create_program(&elf, vm_type)?;
+        let (tx, program_rx) = oneshot::channel();
+        self.writer_tx
+            .blocking_send((
+                Write::Program { vm_type, program_id: program_id.clone(), program_bytes },
+                Some(tx),
+            ))
+            .expect("writer channel broken");
+
+        // Program write will complete after the elf, so we wait on that before
+        // returning to the user.
+        let _ = program_rx.blocking_recv();
 
         Ok(program_id)
     }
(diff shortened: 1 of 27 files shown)


crates/coproc/src/writer.rs
```diff
@@ -52,6 +53,15 @@ pub enum Write {
         /// ELF.
         elf: Vec<u8>,
     },
+    /// Write an program to the program table.
```
emostov, line 56:
```suggestion
    /// Write a program to the program table.
```

crates/coproc/src/intake.rs
```diff
@@ -263,11 +263,25 @@ where
...
+        self.writer_tx
+            .blocking_send((
+                Write::Program { vm_type, program_id: program_id.clone(), program_bytes },
+                Some(tx),
+            ))
+            .expect("writer channel broken");
+
+        // Program write will complete after the elf, so we wait on that before
+        // returning to the user.
+        let _ = program_rx.blocking_recv();
```
Avi-D-coder, line 284: Is it possible for this ever to be `None`?

crates/coproc/src/pool.rs
```diff
@@ -131,16 +131,17 @@ where
...
 
         while let Ok((mut job, reply_tx)) = pool_rx.recv() {
-            let elf_with_meta = match Self::get_elf(&db, &mut job, &metrics, writer_tx2.clone()) {
-                Ok(elf) => elf,
-                Err(_) => continue,
-            };
+            let program_with_meta =
+                match Self::get_program(&db, &mut job, &metrics, writer_tx2.clone()) {
+                    Ok(elf) => elf,
+                    Err(_) => continue,
```
Avi-D-coder, line 137: `continue` feels wrong to me.
Are we sure it's okay o not handle a `PoolMsg`?

Avi-D-coder: Not sending anything back in the oneshot seems like it would cause a bug.

emostov: Good point, will figure something out. LMK if you have thots on best practice

crates/coproc/src/intake.rs
emostov: Not that I know of. But we can return an error just in case

crates/coproc/src/pool.rs
Avi-D-coder: I think just sending an error back over the oneshot is fine then handling it there.
It'd be good if the caller can see there was an error when they check the status, but if the status does not exist I guess we can assume it's an error.

Avi-D-coder (approved):

PR #497, opened 2025-02-04
feat: improve error message for non-allowed transactions

### LINKED ISSUE #471: ivm-exec: create a custom RPC error for transactions denied by allow list
I think it works, my only concern is just that it will confuse users.
Maybe we can create an issue for longer term to create a custom error? probably not a mainnet blocker, but it would be nice eventually

_Originally posted by @lucasege in https://github.com/InfinityVM/InfinityVM/pull/452#discussion_r1922773261_

emostov:
Closes #471

### DIFF
--- a/bin/exec/src/pool/validator.rs
+++ b/bin/exec/src/pool/validator.rs
@@ -749,7 +749,7 @@ pub struct IvmTransactionValidator<Client, Tx> {
     ivm: Arc<IvmOnlyTransactionValidatorInner>,
 }
 
-/// Transaction validation error indicating that the sender is and recipient is not part of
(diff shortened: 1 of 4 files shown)


bin/exec/src/pool/validator.rs
```diff
@@ -748,6 +749,27 @@ pub struct IvmTransactionValidator<Client, Tx> {
     ivm: Arc<IvmOnlyTransactionValidatorInner>,
 }
 
+/// Transaction validation error indicating that the sender is and recipient is not part of
```
emostov, line 752:
```suggestion
/// Transaction validation error indicating that the sender and recipient is not part of
```

PR #499, opened 2025-02-05
chore: Update suzuka allow list to reflect regenerated network

twoshark:
# What 

Suzuka-0 was regenerated and that included all wallets. This updates the allow list to reflect the current wallets.


emostov:
Closed in favor of #500

PR #502, opened 2025-02-09
fix: Use `ReentrancyGuardUpgradeable`

MaanavKhaitan:
# What

- Use the upgradeable version of `ReentrancyGuard`


emostov (commented):
Don't totally understand what this is doing, but seems reasonable

lucasege (approved):

PR #503, opened 2025-02-09
fix: Reduce storage gap in `JobManager`

MaanavKhaitan:
# What

Closes https://github.com/Enigma-Dark/Infinity/issues/1

- Reduce storage gap from 50 to 46 slots so it adds up to 50
- Fix forge tests


emostov (approved):
Seems legit

PR #504, opened 2025-02-09
fix: Add storage gaps in base coproc contracts

MaanavKhaitan:
# What

- Adds storage gaps in the base coproc contracts like `SingleOffchainSigner` and `StatefulConsumer`

### DIFF
--- a/contracts/src/coprocessor/Consumer.sol
+++ b/contracts/src/coprocessor/Consumer.sol
@@ -18,6 +18,9 @@ abstract contract Consumer is
     
     mapping(bytes32 => JobInputs) internal jobIDToInputs;
 
+    // Storage gap for upgradeability.
+    uint256[47] private __GAP;
+
     function initialize(address initialOwner, address jobManager, uint64 initialMaxNonce) public virtual onlyInitializing {
         _transferOwnership(initialOwner);
         _jobManager = JobManager(jobManager);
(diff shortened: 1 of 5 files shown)


contracts/src/coprocessor/Consumer.sol
```diff
@@ -18,6 +18,9 @@ abstract contract Consumer is
     
     mapping(bytes32 => JobInputs) internal jobIDToInputs;
 
+    // Storage gap for upgradeability.
+    uint256[47] private __GAP;
```
emostov, line 22: Is it normal to add a gap to an abstract contract? I would assume the actual gap would be dependent on the actual contract

MaanavKhaitan: Yeah, I saw some of the abstract contracts in the OpenZeppelin repo also had storage gaps. Seems like a standard thing to do

emostov (approved):

contracts/src/coprocessor/Consumer.sol
maarcweiss: You need to add the gap on the abstract contract and not in the mean contract actually

MaanavKhaitan: we have multiple abstract contracts like `Consumer` and `OffchainRequester`, and the main contract can inherit both of them. In that case, which abstract contract should we add the gap on? Should we add the gap in both abstract contracts?

maarcweiss: So if `Consumer` and `OffchainRequester` are only meant to be inherited together by ClobConsumer and not inherited separately in a different main contract then you can add it in `OffchainRequester`, otherwise, if they are meant to be re-used on other main contracts separately in the future, you can leave both

PR #508, opened 2025-02-15
fix: Only storage gap in abstract contract

MaanavKhaitan:
# What

- Removes storage gap from contracts other than the abstract contract `Consumer.sol`

# Testing

`forge test`: Passes


emostov (approved):

maarcweiss (commented):
fixed

PR #522, opened 2025-02-23
chore: regensis suzuka

emostov:


twoshark (approved):
Thank you!

bin/exec/src/config/mod.rs
```diff
@@ -133,50 +133,32 @@ impl IvmConfig {
...
+                    address!("0xcc1e4f683ab0eef001d2dd6c5c87c07a15c71f70"), // validator-2
+                    address!("0xb6da9b28619412d728be31643ee10b1001c7d463"), // validator-3
                 ];
 
                 let mut fork0 = IvmTransactionAllowConfig::deny_all();
                 let mut priority_senders = HashSet::new();
-                for addr in addresses0 {
+                for addr in addresses_sender0 {
                     fork0.add_sender(addr);
                     fork0.add_to(addr);
```
emostov, line 156: Adding to on these addresses for flexibility, but in the future we will just use the deployed contract addresses as valid to and these admin addresses will only be senders

PR #523, opened 2025-02-23
fix: Remove coproc default feature for tracing log level

dist1ll:
Having this feature enabled by default causes tracing's feature flag `release_max_level_info`  to be active for all binaries, if built from the workspace root. The reason for this is [feature unification](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification), coupled with the fact that tracing's flags are technically not additive. 

To avoid this footgun, we should specify the minimum log level via cargo. Example for a release build with info and above:

```
cargo build --bin ivm-exec --release --features min-info-logs
```

### DIFF
--- a/.github/workflows/docker-coproc.yml
+++ b/.github/workflows/docker-coproc.yml
@@ -24,8 +24,8 @@ env:
 
 jobs:
   docker-release:
-    name: Docker release to Google Artifact Registry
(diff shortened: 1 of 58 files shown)


emostov (approved):
Should we also remove the default for reth?

dist1ll:
@emostov reth already doesn't have that default.

PR #524, opened 2025-02-24
chore: add deployed contracts to suzuka

emostov:
Changes:
- Remove senders from allowed `to` field
- Add new contracts to allowed `to` field


MaanavKhaitan (approved):
LGTM

PR #527, opened 2025-02-25
chore(coproc): check next nonce after offchain job relay 

emostov:


lucasege (approved):
Looks good, this would probably have interesting implications for testing, i guess ideally this doesn't break any of our existing tests (seems good) but we could also add a (kinda pedantic) new test when targeting an EOA vs. a contract if we wanted to recreate today's case

emostov:
> Looks good, this would probably have interesting implications for testing, i guess ideally this doesn't break any of our existing tests (seems good) but we could also add a (kinda pedantic) new test when targeting an EOA vs. a contract if we wanted to recreate today's case

Right now it is just a log, so not sure there is much to check

PR #532, opened 2025-02-25
chore: Add `getJobManager()` to Consumer contract

MaanavKhaitan:
# What

- Adds `getJobManager()` to `Consumer.sol`


emostov (approved):
Nice ty

